PR #3, opened 2017-11-21
Enable the state.highstate button for a particular minion, some styling adjustments

maerteijn:


maerteijn:
PS, I really liked your approach of this project: Just plain ES6 with no npm / webpack / other libraries. And it works in recent browsers, it is simple and you can adjust things without all kinds of hassle 👍

oliverdunk:
Thanks for this @maerteijn! I don't use Salt much anymore, hence the lack of commits, but I'll try to look at this and get it merged for you shortly.

By the way, thank you for the kind words about my approach! Frameworks are great but I'm personally a huge fan of Vanilla JS when it's practical.

oliverdunk:
Hi Martijn,

Thank you for all of these.

Would you be interested in taking over ownership of the repository? I don't have much time to spend on this and after so long, you are probably now more familiar with the code than I am!

maerteijn:
Hi @oliverdunk ! Yes sure! We will extend the tool the following months so that would be nice.
I think you can transfer the ownership in the github menu somewhere.

oliverdunk:
I have initiated a transfer request. All I ask is that you keep the licence which is currently in place. Some credit in the README would be nice too :)

Thank you so much for your support! I'm excited to see where you take this.

maerteijn:
@oliverdunk Yes, I'll keep the license in place and of course your name will get a prominent place in the credits :-)

Thank you for the transfer!

PR #16, opened 2018-05-19
Logout

erwindon:
addition of a logout button
logs out both remotely (api call "/logout") and locally (remove cookie)
returns to the login screen on logout
needed slight refactor of existing "manualrun" button, both plumbing and visualization (but not implementation)


maerteijn:
Thanks!

I was not such a fan of the button layout. So I changed it to a subtile link like here:

layout

PR #18, opened 2018-05-29
Sessiontimeout

erwindon:
Solution for session timeout. Does not cover all session timout effects, but at least the page-refresh of the home screen returns to the login screen.

Also made the username/password fields readonly on succesful login. The still blinking cursor is then prevented.

### DIFF
--- a/saltgui/static/scripts/api.js
+++ b/saltgui/static/scripts/api.js
@@ -176,6 +176,13 @@ class API {
   }
 
   _onFetchResponse(response, resolve, reject) {
+    if(response.status == 401) {
+      // sesion has expired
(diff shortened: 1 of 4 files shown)


maerteijn:
Thanks again!

I would say: change the read-only attribute to disabled to have a more clear effect:
```javascript
document.querySelector("#username").disabled = true;
document.querySelector("#password").disabled = true;
```

I already fixed the` background-color` for inputs in the css in the master branch

erwindon:
Cherry-picked your CSS changes, and applied your suggestion for disabling the login/password fields.
Reconfirming pull request...

PR #19, opened 2018-05-30
Keymanagement

erwindon:
Implemented key-management api calls.
Added relevant commands for key manipulation with each minion/key.
Re-query the minion list after key manipulation commands.
Some practical changes to the CSS to allow multiple right-aligned commands.
Hide the "Inactive Minions" title when there are none.
Layout/styling suggestions welcome...


maerteijn:
The functionality looks awesome! I'll review next Monday.

erwindon:
Note that the key-manipulation commands will move to their own page as soon as #20 is realized.

saltgui/static/scripts/api.js
```diff
@@ -74,6 +74,14 @@ class API {
...
     manualRun.style.display = isShowing ? "none" : "block";
     document.body.style["overflow-y"] = isShowing ? "scroll" : "hidden";
+
+    // test whether the command may have caused an update to the list
+    // the user may have altered the text after running the command, just ignore that
+    var command = document.querySelector(".run-command #command").value;
+    var output = document.querySelector(".run-command pre").innerHTML;
+    if(isShowing && command.startsWith("salt.wheel.key.") && output != "Waiting for command...") {
+      location.reload(); 
+    }
```
maerteijn: Hmmm, I'm not really happy with this solution. Wouldn't it be better to store something in a state which decides to reload the page or not?

saltgui/static/scripts/api.js
```diff
@@ -133,13 +141,33 @@ class API {
...
+      // See https://docs.saltstack.com/en/latest/ref/wheel/all/salt.wheel.key.html#salt.wheel.key.delete
+      params.client = "wheel";
+      params.fun = functionToRun.substring(11);
+      params.match = target;
+    } else {
+      params.client = "local";
+      params.fun = functionToRun;
+      params.tgt = target;
+      if(args.length !== 0) params.arg = args.join(" ");
+    }
```
maerteijn: I see a lot of code duplication here. I think this can be done a lot cleaner

saltgui/static/scripts/routes/home.js
```diff
@@ -181,8 +260,26 @@ class HomeRoute extends Route {
...
+      this._runDeleteKey(hostname, evt);
+    });
+    element.appendChild(deleteButton);
+
+    var acceptButton = this._createDiv("run-command-button", "Accept &#9658;");
+    acceptButton.addEventListener('click', evt => {
+      this._runAcceptKey(hostname, evt);
+    });
+    element.appendChild(acceptButton);
+
```
maerteijn: Also here I see a lot code duplication of creating the buttons here. I think you should create some helper functions for this.

maerteijn:
A comment about the lay-out: with all these buttons it will become a mess. Also if you move the key management stuff in a separate page:

saltgui ui

I suggest that we need a new kind of control for this (a dropdown like control), or that we move the selection of the command inside the popup, so there is just one button in the overview:

saltgui ui suggestion

saltgui/static/scripts/api.js
erwindon: The user has the freedom to change commands. That is why the actual command that is used needs to be tested. I do intend to remove this crude function and replace it with a mechanism based on the information from either "GET /events" and/or "GET /ws", In that case the server is pushing changes back to the client. The result is a dynamically updating screen. Also when actions are performed outside SaltGUI. My proposal is to leave this code until that is done.

saltgui/static/scripts/api.js
erwindon: Agreed, working on it...

saltgui/static/scripts/routes/home.js
erwindon: Agreed, working on it...

erwindon:
A next pull request will include a menu bar and splitting functionality across pages. This will reduce the number of commands on the main page back to the original "Sync state ►". My screen is a bit wider so it was no problem for me yet. Let me see whether I can brew something for this. Probably something like https://www.w3schools.com/howto/howto_css_dropdown.asp

saltgui/static/scripts/api.js
erwindon: Separated the client/fun/match handling from the rest, since these were present 3 times.

erwindon:
The dropdown menu worked out nicely (if I may say so).
Added that to this PR because it reduced layout problems that were otherwise introduced here.

saltgui/static/scripts/routes/home.js
erwindon: Simplified this code.

saltgui/static/scripts/api.js
maerteijn: Ok, then we will leave it for now.

saltgui/static/scripts/routes/home.js
```diff
@@ -75,6 +90,51 @@ class HomeRoute extends Route {
...
+    var menuButton = this._createDiv("menu-dropdown", "&#9658;");
+    menuDropdown.appendChild(menuButton);
+    var menuDropdownContent = this._createDiv("menu-dropdown-content", "");
+    //menuDropdownContent.appendChild(this._createDiv("run-command-button", "x1"));
+    //menuDropdownContent.appendChild(this._createDiv("run-command-button", "x2"));
+    //menuDropdownContent.appendChild(this._createDiv("run-command-button", "x3"));
+    menuDropdown.appendChild(menuDropdownContent);
+    element.appendChild(menuDropdown);
+    return menuDropdownContent;
+  }
```
maerteijn: Ok, I really like the menu popup dropdown, it makes much more sense like this! Personally I would make the font a bit smaller so it fits a bit better with the rest:

menu

As for the code, I would suggest to create a new class for the menu and create an instance of it in the constructor (like this: https://github.com/maerteijn/SaltGUI/blob/master/saltgui/static/scripts/router.js#L4), so you have the menu code separated from the rest (otherwise we will create one big monolithic class which will be hard to maintain)

saltgui/static/scripts/routes/home.js
```diff
@@ -75,6 +90,51 @@ class HomeRoute extends Route {
...
+
+  _addMenuItemSyncState(hostname) {
+    var highStateButton = this._createDiv("run-command-button", "Sync&nbsp;state");
+    highStateButton.addEventListener('click', evt => {
+      this._runHighState(hostname, evt);
+    });
+    return highStateButton;
+  }
+
```
maerteijn: I see a lot of duplication here. If you would create a separate class of the menu you could create a private method on this class as a helper to have less duplication.

saltgui/static/scripts/api.js
```diff
@@ -133,13 +141,31 @@ class API {
...
-    var params = {
-      tgt: target,
-      fun: functionToRun,
-      client: "local"
-    };
+    var params = {}
+
+    if(functionToRun.startsWith("salt.wheel.key.")) {
+      params.client = "wheel";
+      params.fun = functionToRun.substring(11);
```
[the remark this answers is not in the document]

saltgui/static/scripts/api.js

saltgui/static/scripts/routes/home.js
erwindon: The font enlargement has been removed

erwindon: The dropdownmenu system has been refactored into a separate class

erwindon:
(unless I forgot something) the requested changes have all been made.

maerteijn:
👍

PR #22, opened 2018-06-06
Cleanup

erwindon:
Cleaned up in 2 categories
1) solved all warnings as produced bij jslint with setting "esversion: 6"
2) removed unused parameters in 3 constructor calls

### DIFF
--- a/saltgui/static/scripts/api.js
+++ b/saltgui/static/scripts/api.js
@@ -67,8 +67,7 @@ class API {
 
   _toggleManualRun(evt) {
     var manualRun = document.querySelector("#run-command-popup");
-    var isShowing = manualRun.style.display !== "none"
-      && manualRun.style.display !== "";
+    var isShowing = manualRun.style.display !== "none" && manualRun.style.display !== "";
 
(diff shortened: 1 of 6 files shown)


maerteijn:
Thanks. When I have time I will add a linter which checks the pull requests for ES6 linter warnings / errors.

maerteijn:
@erwindon Can you show me the exact settings / config of jslint you used? Then I can add it as post-commit hook.

erwindon:
$ cat ~/.jshint 
{
"esversion": 6
}`

erwindon:
and commandline:
`jshint --config ~/.jshint --exclude-path saltgui/static/scripts/JSON.sortify.js`
The excluded path is probably also possible in the jshint file.
JSON.sortify will be introduced later when we want JSON.stringify functionality, but then with sorted keys for dictionaries.

maerteijn:
@erwindon Thnxm added this in #28 

You can add the exclude when needed  here https://github.com/maerteijn/SaltGUI/blob/master/package.json#L9

Issue #20, opened 2018-05-31
Menu bar in SaltGUI


erwindon:
I want to split the recent additions for keys-management to a separate page.
First page would then be the traditional minions overview.
Second page would then be the recent keys-management page.
And there would be room for a few more pages as needed.
So, a simple menu bar is needed to select between the pages.

Could look like this:

![saltgui_menu](https://user-images.githubusercontent.com/3663742/40807662-3d48c36a-6525-11e8-84ff-222e4dbb5aca.png)

any objections, improvements, etc?

maerteijn:
This would indeed be a pretty nice improvement. The styling of the menu looks fine to me like this!

maerteijn:
@erwindon , there is something here that I should mention which I don't know (yet) how to solve, so let's put my thoughts here and see what you think.

Until now, SaltGUI was also responsive, and did a pretty good job with that:

<img width="334" alt="responsive" src="https://user-images.githubusercontent.com/121401/43128756-1996b250-8f33-11e8-9995-999ed26a298b.png">

I like that, because I'm pretty sure this application is (or will be) used on phones/tablets as well.

But with the introduction of the menu, this will become a lot more difficult without any kind of hamburger menu or so:

<img width="320" alt="responsive_with_menu" src="https://user-images.githubusercontent.com/121401/43128798-2deb4dc4-8f33-11e8-8737-cfd10892e74f.png">

This will become worse when we add more menu-items. Also, the css files become bigger and bigger, so fixing this will be harder with each improvement we add.

And then we are not talking about all the functionality you have waiting in your feature branches.

And I don't want SaltGUI to become a hard, unmaintainable project.

So, or we should fix the css for this. But I'm not a super css hero actually. I'm doing this project for fun, to learn, and because I promised the original author to take good care of SaltGUI. Doing css stuff which has been done 100 times before with bootstrap / foundation / bulma feels stupid to do.

So maybe we should think about introducing a css framework here (I like bulma for it's simplicity) and do a complete restyling  with that. Which also means we can use all it's components which are pretty nice. Which also means it takes some time to do this and that it will take more time for your other features to be merged.

We could also do some quick hacks first: disable the menu on a small resolution and some css fixes here and there and then move on to the bulma approach. Which makes more sense when you think the scrum way.

So now I put this all on the table, your thoughts please :)

erwindon:
[...]

PR #24, opened 2018-06-09
Addition of menu bar; separation of minion vs keys

erwindon:
This PR introduces the menubar as proposed in #20.
The first 2 pages are actually more-or-less the original home page, but split by the minion vs keys functionality.
The original (single) home page is renamed to "minions" page. This one now again contains only entries for active minion-keys. A second page contains the "keys" functionality. This lists all known keys, (i.e.: accepted, rejected, unaccepted and denied keys). Each page shows only the commands (in the dropdown menu) that match the purpose of the page.
The menu-bar code allows pages (such as job details) without corresponding menu-item.
The menu-bar keeps the menu item for the active page highlighted.


maerteijn:
My first impression is: Looks good! I think the majority of the pull request is about moving code around because you added a extra page right?

PS, could you rebase this branch on master (conflicts)

erwindon:
> I think the majority of the pull request is about moving code around because you added a extra page right?
Yes, and the fact that there are now multiple pages.

The branch has been rebased as needed/requested.

On re-test, I noticed that navigating between the pages still seems to have a timing problem. When you navigate while one/more api-calls is/are still running, then screen becomes empty. It can then only be restored using full page refresh.
I'll look into that first...

erwindon:
The actual problem is cache-control. Under certain circumstances, an API call returns the HTML from the browser cache. That is now prevents.
The PR is valid again.

saltgui/static/scripts/routes/keys.js
```diff
@@ -0,0 +1,155 @@
...
+    container.appendChild(element);
+  }
+
+  _addPreMinion(container, hostname) {
+    var element = document.createElement('li');
+
+    element.appendChild(Route._createDiv("hostname", hostname));
+
+    var pre = Route._createDiv("unaccepted", "unaccepted");
+    pre.id = "status";
```
maerteijn: Adding multiple elements with the same id is not allowed in the dom. So you should add a class here. See also my general comments about this pull request

saltgui/static/stylesheets/main.css
```diff
@@ -71,6 +71,22 @@ h1 {
   cursor: pointer;
 }
 
+.menu_item {
+  display: inline-block;
+  padding-left: 20px;
+  font-size: 20px;
```
maerteijn: Can you change this to `18px` ?

maerteijn:
I think this is a change for the better (and the future) but there are some elements I'm not so keen on (yet). Let's take the following example:

key-duplicates

- When you reject a key, the page will quickly reload reload but still shows the same key as accepted here. This is not the case when you completely reload the page. 
- I don't see why it displays 'rejected' next to ``8be47cba35e2`` when it's already listed under 'rejected keys'. See the comment here below.
- The layout with different fields (os description in the accepted keys list, no description in the others) becomes a bit of a mess. Wouldn't it be a better idea to do something like:
menu-suggestion

Note that you use the same class for the `status` and the `os` display to make sure the button alignment stays a bit sane.

saltgui/static/scripts/dropdown.js
```diff
@@ -17,6 +18,7 @@ class DropDownMenu {
     var button = Route._createDiv("run-command-button", title);
     button.addEventListener('click', evt => callback(evt));
     this.menuDropdownContent.appendChild(button);
+    this.menuDropdown.style.display = "block";
```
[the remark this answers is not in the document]
erwindon: done

saltgui/static/stylesheets/main.css
erwindon: done (I assume you meant the font-size)

erwindon:
Strange, the page reload after a successful key-management api call does a `location.reload()`, which is the same a complete page reload. The problem seems to be (again) a timing error. We use 2 api calls: 1 to get the keys, and 1 to get the grains info. In your case, the effect can only be reproduced when the grains info returns faster than the keys info. Because the grains info is still returned, an item is added to the "accepted" block. Because the keys info says the key is rejected, an item is added to the "rejected" block. therefore ending with 2 items on the screen. I'll make the keys information leading on page "keys" and remove any accidental successful grains information.
please wait...

erwindon:
I have a lot of offline minions (average 50%) so that effect is less visible here. The denied/unaccepted/rejected sections are mostly empty. The other pages that I'm proposing also have the same layout with both status/ip and os. Can we postpone this layout change to a bit later?

maerteijn:
Ok, I'm willing to postpone the layout changes later. The only thing that really is a concern for me is this:

div layout

The accepted keys section has three columns, the other one two. Could you at least have the same amount of columns (with the same css width etc) to have the buttons aligned equally?

erwindon:
Finally, I understand... In the resolution that I used, the command buttons were floating-right. But not in the lower resolutions. Then indeed all columns must be present (easy fix).

erwindon:
I think I've updated all the requested changes.
Note that I treated it as the original PR, i.e. it is a PR only for branch `pages3`.
Our common effort on feature `hamburger` is not included here.
That branch now needs a rebase and should be in its own merge (with or without a PR).

maerteijn:
Fantastic work with the fixes, just tested and looks fine!

PR #27, opened 2018-07-17
Added support for generic EAUTH

erwindon:
Currently SaltGUI supports only PAM as authentication method.
This PR adds support for the other authentication methods.
A dropdown box is added to the login screen.
When nothing is entered, PAM is used as before.
The user's choice is persisted in the browser's local storage, so that on the next login, the default is usually already correct.
Regression tested with PAM implicitly selected. Tested with PAM explicitly selected. Tested also with AUTO, FILE, SHAREDSECRET and YUBICO.

### DIFF
--- a/README.MD
+++ b/README.MD
@@ -58,6 +58,9 @@ This excellent frontend is originally written by [Oliver Dunk](https://github.co
 
 ## Changelog
 
+## 1.2.0 (2018-07-17)
+- Added support for alternative eauth methods [#23](https://github.com/maerteijn/SaltGUI/issues/26) (erwindon)
+
 ## 1.1.0 (2018-07-16)
 - Shows inactive minions as well (erwindon)
 - Switch to a more reliable grain indicating the ip-number (erwindon)
--- a/README.MD
+++ b/README.MD
@@ -58,9 +58,6 @@ This excellent frontend is originally written by [Oliver Dunk](https://github.co
 
 ## Changelog
 
-## 1.2.0 (2018-07-17)
-- Added support for alternative eauth methods (erwindon)
-
 ## 1.1.0 (2018-07-16)
 - Shows inactive minions as well (erwindon)
 - Switch to a more reliable grain indicating the ip-number (erwindon)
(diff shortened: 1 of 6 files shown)


saltgui/index.html
```diff
@@ -27,9 +27,32 @@ <h1>SaltGUI</h1>
...
+              <option value="sharedsecret">sharedsecret</option>
+              <option value="yubico">yubico</option>
+            </optgroup>
+            <optgroup label="experimental" />
+              <option value="django">django</option>
+              <option value="keystone">keystone</option>
+              <option value="ldap">ldap</option>
+              <option value="mysql">mysql</option>
+              <option value="pki">pki</option>
+              <option value="rest">rest</option>
```
maerteijn: Experimental means untested / assumptions. I prefer not to have this in the codebase: just add new login types when they are fully tested or are a requirement. So we should add `ldap` as well and check with @babs if this works.

saltgui/static/scripts/api.js
```diff
@@ -113,6 +113,12 @@ class API {
       eauth: "pam"
     };
 
+    // overrule the eauth method when one is selected
+    var type = document.querySelector("#login-form #eauth");
+    if(type.value !== "default")
+      params.eauth = type.value;
```
maerteijn: Could you use brackets or change this as one line?

README.MD
```diff
@@ -58,6 +58,9 @@ This excellent frontend is originally written by [Oliver Dunk](https://github.co
 
 ## Changelog
 
+## 1.2.0 (2018-07-17)
+- Added support for alternative eauth methods (erwindon)
+
```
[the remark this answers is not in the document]
erwindon: that was a wrong/silly attempt to save you some work, done

saltgui/static/scripts/api.js

saltgui/index.html
erwindon: agreed, done

erwindon:
all requested changes are done

Issue #26, opened 2018-07-17
Config file


babs:
Hi,

It would be great to have a config file somehow, every time I want to update, I have to patch the code to switch eauth from pam to ldap :/
Is would be nice to store it somewhere once and for all.

erwindon:
Can you please show what the exact changes are?
Is it only the {{eauth}} value in {{api.js|}?
And anything else than the changes in section {{external_auth}} in {{/etc/salt/master}}?

The allowed values seem to come from {{/usr/lib/python3.4/site-packages/salt/auth}}.
Just tested it with "auto". That seems to work fine.

My proposal would be to provide a dropdown box with all the known eauth methods and use the selected value. The selected value can be stored in a cookie, so that you do not have to update the dropdown box each time.

erwindon:
With a bit of experimentation, it is now clear.
I've issued PR #27 for this.

maerteijn:
Hi @babs 

I just merged #27  which allows you to select the login method. Could you test this with your LDAP setup to see if it works for your scenario?

babs:
it works, but LDAP is not experimental :D

erwindon:
LDAP was experimental because no-one had tested it yet with SaltGUI. I did not setup (or find) a working LDAP server to do that. Now that you have tested it, I've updated the GUI and issued a PR for that.

babs:
oh, Ok :D

PR #29, opened 2018-07-20
promoted ldap because it was tested ok

erwindon:
This concludes #26


maerteijn:
I squashed the pull request because it was so small. I won’t do it anymore because of you having 10+ branches ready for the future, then commits will ‘disappear’ on your side.

PR #33, opened 2018-07-28
Added documentation commands to runbox

erwindon:
Added a dropdown menu just after the `Command` field from the command box.

The menu contains several documentation commands:
1) using the internal documentation system, actually submitting the command `sys.doc` with the already typed command as argument. When the user did not fill in a target yet, assume `*`. Select only the first answer just like the `salt` commandline does for `sys.doc` commands. Updated output format to show the help topic instead of the minion name; and showing only the help text instead of the technical output.
2) using external URLs to the saltstack website:
    * display list of available module categories. menu item always shown;
    * display list of available modules in category. always shown, default category is `execution`;
    * display list of available commands in module. shown when command field contains at least a module name;
    * display details of the command. same as 2c, but scrolls down to the command, shown when .command field contains at least a module name and a command name).

(part 2 is no longer part of this PR)

Note that the menu system was extended to cope with dynamic visibility and dynamic text of menu items.


maerteijn:
Can you rebase this?

erwindon:
Just rebased it. But I still need to minimize the changeset...

erwindon:
It is now ready for review/merge.

erwindon:
rebasing now...

erwindon:
rebase complete and retested.

erwindon:
Travis is failing after I accidentally pushed my branches into your repository; and then removed those again. Travis still thinks they exist...

erwindon:
I accidentally found the salt commands `runners.doc.wheel` and `runners.doc.runner`, which provide the online documentation for `wheel` and `runner` commands. Updated SaltGUI to use these functions when requesting documentation for `wheel` or `runner` commands.
Note that support to actually run `runners` commands is added in a different branch (future PR).

maerteijn:
Will need a bit more time to review it. Hopefully I can do that this week.

erwindon:
please wait for rebase (which should solve the issues)

erwindon:
rebased (and squashed); please continue the review

erwindon:
1) In another PR/branch, I'm unifying the code that produces the output for running commands and for viewing historic jobs. the documentation formatter is part of that. I'll move those parts to a separate file.
2) on my system it was just (really) slow, hence the additional warning. I'll remove that menu item when the command field is empty. I'd like to keep it a dropdown menu because that explains what will be executed. Just need to think of a good title.
3) I'll drop the external documentation part from this PR. I'll keep that one on its own feature branch. Unless something fundamentally changes it will not return. I added that part to overcome the missing runner/wheel documentation in `sys.doc`. This later had another solution in the form of `runners.doc.*`. I did not then remove the online part because it is more readable and it allows to be expanded to other areas such as (salt) states.

erwindon:
Everything is reorganized.
New class "Documentation" (obviously) handles specifics for getting and showing documentation.
New class "Output" knows how to represent a salt output. Current knowledge is about the following output formats: errors, documentation and regular. In a future PR, the jobs viewer will use this class too; and the "state" output format will then also be supported (only) in this class.

saltgui/static/scripts/api.js
```diff
@@ -6,8 +6,17 @@ class API {
     this._callMethod = this._callMethod.bind(this);
     this._fetch = this._fetch.bind(this);
     this._getRunParams = this._getRunParams.bind(this);
+    this._manualRunMenuSysDocPrepare = Documentation._manualRunMenuSysDocPrepare.bind(this);
+    this._manualRunMenuSysDocRun = Documentation._manualRunMenuSysDocRun.bind(this);
```
maerteijn: Much much better to have this in a separate class.

saltgui/static/scripts/api.js
```diff
@@ -33,12 +42,26 @@ class API {
...
+    // cut/paste do not work everywhere
+    document.querySelector("#command")
+      .addEventListener('cut', this.menu.verifyAll);
+    document.querySelector("#command")
+      .addEventListener('paste', this.menu.verifyAll);
+    // blur/focus should not be needed but are a valueable fallback
+    document.querySelector("#command")
+      .addEventListener('blur', this.menu.verifyAll);
+    document.querySelector("#command")
+      .addEventListener('focus', this.menu.verifyAll);
```
maerteijn: Would it be nice if tthese specific events would be defined in the Document class and that you just call:
```javascript
Document._registerEventListeners();
``` 

?

saltgui/static/scripts/api.js
```diff
@@ -49,32 +72,20 @@ class API {
...
-    outputContainer.innerHTML = "";
+  _onRunReturn(command, data) {
+    let response = data.return[0];
 
-    for(var i = 0; i < hostnames.length; i++) {
-      var hostname = hostnames[i];
+    let outputContainer = document.querySelector(".run-command pre");
 
-      var output = response[hostname];
+    Output.addOutput(outputContainer, response, command);
```
maerteijn: I also like the Output class!

saltgui/static/scripts/api.js
removed these 2 lines.
replaced with explicit parameter instead of binding to another class.
one of these functions did not even need `this`.

saltgui/static/scripts/api.js
erwindon: I want to keep the menu creation in `API.js` because the menu may someday contain items that are not documentation related.
I want to keep the event registration to `menu.verifyAll` in `API.js` because that code is not documentation related. i.e. it has no technical connection to the documentation logic.
Did move the registration of the menu item to the `Documentation` class.

erwindon:
ping... :-)

maerteijn:
@erwindon pong! Tomorrow I'll have time to review it again.

PR #34, opened 2018-07-29
Added better command parsing

erwindon:
The current command parsing has some disadvantages.
This PR improves the following:
* The first word of a command line was treated as the command and the rest of the line as a single string argument. This was sufficient for simple usage. Now the remaining text is divided into separate arguments. The parser is smart and recognizes the python data-types such as string (start with `"`), lists (start with `[`) or dictionary (start with `{`). Arguments that look like integers or floats are treated as such. Also the special constants `none`, `null`, `true` and `false` are recognized. Unrecognised words are strings, so quotes are still not required for individual arguments.
* Named parameters are now also recognized. This also eliminates the special cases for the key management commands. So that is cleaned up too.
* A helpful error message is shown when the command-line cannot be parsed.


erwindon:
The actual parsing happens in JSON.parse everything else is just preparation, checking and management.
We recognize only 3 types: tokens that start with a known character (`[{"); tokens that match a known regexp; and other tokens. This syntax will not be extended.
I've split code in 2 parts: the application logic and the actual parser
(text to be completed)

saltgui/static/scripts/api.js
```diff
@@ -153,40 +164,190 @@ class API {
...
+      } else if(toRun[0] === '\'') {
+        endChar = '\'';
+        objType = "single-quoted-string";
+      }
+
+      var value;
+      if(endChar && objType) {
+        // The string starts with a character for a known JSON type
+        var p = 1;
+        while(true) {
```
maerteijn: Isn't there something you can use instead of `while (true)` ? When you make a mistake you will have an infinitive loop.

erwindon: The loop only repeats from the `continue` in error handler `catch(err)`.
All other paths have a `return` or a `break`.
The advancement of the position-counter `p` in that exception handler makes sure that the loop always ends (a infinite strings do not exist).

erwindon:
The function is moved to file `scripts/parsecmdline.js`.

PR #37, opened 2018-08-03
Added functional and unit tests

maerteijn:


maerteijn:
@erwindon could you review this pull request before merging? And I have one request for you, see:

https://github.com/maerteijn/SaltGUI/blob/nightmarejs/tests/unit/parsecmdline.js#L21

Could you add some more assertions there? :)

oliverdunk:
This is really cool!

erwindon:
Debian 9:
```
$ bash runtests.sh      
ERROR: There are no scenarios; must have at least one.
Removes stopped service containers.

By default, anonymous volumes attached to containers will not be removed. You
can override this with `-v`. To list all volumes, use `docker volume ls`.

Any data which is not in a volume will be lost.

Usage: rm [options] [SERVICE...]

Options:
    -f, --force   Don't ask to confirm removal
    -v            Remove any anonymous volumes attached to containers
    -a, --all     Obsolete. Also remove one-off containers created by
                  docker-compose run
```

any suggestions?

maerteijn:
@erwindon You have the wrong yarn I think.  See: https://github.com/yarnpkg/yarn/issues/2821 and then the second answer

erwindon:
next one...
```
$ mocha --bail --trace-warnings --reporter spec tests/unit/ tests/functional/
/usr/bin/nodejs: bad option: --trace-warnings
error Command failed with exit code 9.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```
I removed this from `package.json` and now the first test result comes in.

```
/home/erwindon/SaltGUI/saltgui/static/scripts/parsecmdline.js:109
      let str = "";
      ^^^
SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
```
Since this is a complaint about the file we are testing, I guess it is working.

maerteijn:
This sounds like an ancient version of nodejs. Which version do you have installed?

erwindon:
```
$ nodejs --version
v4.8.2
```
Seems I cannot trust the Debian distributions any more.
Now upgraded to node.js 8.11.3 and yarn 2.7.6
Much better now!

erwindon:
Still, I need to use:
version `2` in `docker-compose.xml`
`rm -f` instead of `rm -f -s` in `package.json`
And I'm ignoring a timeout failure in the login test.

maerteijn:
What version of docker-compose do you have installed? 
`docker-compose --version`

erwindon:
that was docker-compose 1.8.0 from the latest Debian distribution
replaced it with docker-compose 1.22.0 using https://docs.docker.com/compose/install/#install-compose
these technical problems now gone and still ignoring the timeout failure on the login test

maerteijn:
I documented the minimal versions of all tools required to run the tests in 8c6e0b7cb6d2b9dec37185b0e50bc92c9e20d5a4

Now the question remains how to run the integration tests. Do you have a graphical environment? If not, you should have `xvfb` installed, a virtual window manager. (`apt-get install xvfb`).

Then in a shell:
```bash
export DISPLAY=':99.0'
Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
```
Or when you develop remotely with ssh you should enable X11 window forwarding so the screen is displayed locally. Also useful for debugging or viewing the tests.

erwindon:
Added 45 testcases with 150 assertions.
Found 1 error: javascript does not support single-quoted strings, which `parsecmdline` previously tried to support (but always fails). The removal of that code was lost during a rebase and the corresponding anti-test was indeed not repeated until now.
Where can I commit the updated/fixed parser?
Where can I commit the updated unit tests? (the unit test breaks when run without the updated parser)

erwindon:
@oliverdunk good to see you are still involved, thanks for your initial effort!

erwindon:
I'm a bit lost. First there was `zombie.js`, which was only network based. Now there is `nightmare.js` which needs a (virtual) display. Am I correct when I understand that that is optional?

tests/functional/login.js
```diff
@@ -0,0 +1,60 @@
...
+  });
+
+  describe('Login', () => {
+  
+    it('we should be redirected to the login page', done => {
+      browser
+        .end()
+        .evaluate( () => { return document; } )
+        .then(function (document) { 
+          assert.equal(document.location.href, 'http://localhost:3333/login');
```
erwindon: Can you use the `const url` here?

package.json
```diff
@@ -17,6 +17,6 @@
   "scripts": {
     "lint": "jshint saltgui/ tests/",
     "wait-for-docker": "node tests/helpers/wait-for-docker.js",
-    "test": "mocha --trace-warnings --reporter spec tests/unit/"
+    "test": "mocha --bail --trace-warnings --reporter spec tests/unit/ tests/functional/"
```
erwindon: Is the `--bail` needed?

README.MD
```diff
@@ -74,6 +74,12 @@ We use the following testing libraries:
 - [mocha](https://https://mochajs.org/), a well documented testing framework for javascript 
 - [chai](http://www.chaijs.com/), the preferred assertion library for testing
 
+You'll need at least:
+- `docker-compose` 1.12 or above
+- `nodejs` 8.11.x or above
+- `yarn` 1.7.0 or above
+
```
erwindon: so that would be "8.11 or above" and "1.7 or above"?

erwindon:
Additional question: in `README.MD` the sample configuration now also uses port 3333. Was that intentional?

README.MD
maerteijn: Yes, small typo

maerteijn:
The 3333 was intentional yes as it’s referred to it further in the readme.

maerteijn:
I dropped zombie.js as it took way too much time to write working tests. It is a bit buggy and there is no up to date documentation. 

Nightmare.js works pretty awesome and is well documented. Tests can be debugged with a debugger which is pretty neat. The only thing is yes, you’ll need a display driver.

maerteijn:
I would put the updated parser and the tests for it here in this branch

tests/functional/login.js
```diff
@@ -0,0 +1,80 @@
+const Nightmare = require('nightmare');
+const assert = require('chai').assert;
+
+const url = 'http://localhost:3333/';
+
+
+
```
oliverdunk: Is there a reason for the extra new lines here?

tests/functional/login.js
```diff
@@ -0,0 +1,80 @@
...
+  let browser = null;
+
+  beforeEach( () => {
+    browser = new Nightmare({ 
+      // to make the typed input much faster
+      typeInterval: 20,
+      // uncomment this to show the browser and the debug window
+      // openDevTools: {
+      //   mode: "detach"
```
oliverdunk: It would be great to have the script take in a parameter which could be used to change this setting. The default should probably be a hidden browser window and it should be shown only if requested.

runtests.sh
```diff
@@ -0,0 +1,25 @@
...
+# start a salt master, two salt minions and saltgui to run tests on
+docker-compose -f docker/docker-compose.yml up -d
+
+# wait until all are up
+yarn wait-for-docker
+
+# run the unittests/nightmare.js functional tests
+yarn test
+
+set +e
```
oliverdunk: Missing newline at the end of this file 😉

saltgui/static/scripts/utils.js
```diff
@@ -1,28 +1,35 @@
...
+      var minutes = Math.round(secondsPassed / 60);
+      return minutes + " minutes ago";
+    }
 
-  if(secondsPassed < 60 * 60 * 24 * 2) {
-    return "Yesterday";
-  }
+    if(secondsPassed < 60 * 60 * 24) {
+      var hours = Math.round(secondsPassed / 60 / 60);
+      return hours + " hours ago";
```
oliverdunk: Could we end up with "one hours" here?

saltgui/static/scripts/utils.js
```diff
@@ -1,28 +1,35 @@
...
-  return "A long time ago, in a galaxy far, far away";
+    if(secondsPassed < 60 * 60 * 24 * 30) {
+      var days = Math.round(secondsPassed / 60 / 60 / 24);
+      return days + " days ago";
+    }
+
+    return "A long time ago, in a galaxy far, far away";
+  }
+  catch(err) {
+    return "It did happen, when I don't know";
```
oliverdunk: We should probably log this error the console?

tests/functional/login.js
```diff
@@ -0,0 +1,80 @@
...
+    });
+    browser
+      .goto(url);
+  });
+
+  describe('Functional tests for login', () => {
+  
+    it('we should be redirected to the login page', done => {
+      browser
+        .end()
```
oliverdunk: Can you explain the purpose of the end here? I'm not sure I get it.

tests/functional/login.js
```diff
@@ -0,0 +1,80 @@
...
+        .type('#username', 'sald')
+        .type('#password', 'sald')
+        .click('#login-submit')
+        .wait('.notice')
+        .end()
+        .evaluate(selector => {
+          return document.querySelector('.notice-wrapper div').textContent;
+        }, selector)
+        .then(function (message) {
+          assert.equal(message, 'Authentication failed');
```
oliverdunk: It's a bit of a shame to have the tests rely on particular text which might change in the future, but I can also see the value in checking the error. Just thought I would say this to spark a discussion.

tests/functional/login.js
```diff
@@ -0,0 +1,80 @@
...
+          done() ;
```
oliverdunk: Should there be a space before this semicolon?

tests/functional/login.js
```diff
@@ -0,0 +1,80 @@
...
+        })
+        .catch(done);
+    });
+
+    it('valid credentials will redirect us to the homepage and hide the loginform', done => {
+      browser
+        .type('#username', 'salt')
+        .type('#password', 'salt')
+        .click('#login-submit')
+        .wait( ()=> {
```
oliverdunk: Missing space after the `()`?

tests/helpers/wait-for-docker.js
```diff
@@ -0,0 +1,19 @@
...
+var resources = {
+  resources: [
+    'http://localhost:3333/',
+  ],
+  timeout: 30000,
+  followAllRedirects: true,
+  followRedirect: true,
+};
+
+waitOn(resources, error => {
```
oliverdunk: Since we only use `waitOn` once, I wonder if we can write our own method to check every few seconds instead? It just seems like there's the opportunity here to avoid a dependency.

tests/unit/utils.js
```diff
@@ -0,0 +1,60 @@
+ const assert = require('chai').assert;
+
+// create a global window so we can unittest the window.<x> functions
+if (!global.window)
+   global.window = new Object({});
+
+require('../../saltgui/static/scripts/utils');
+
+
+describe('Unittests for utils.js', function() {
```
oliverdunk: Some tests for potential mistakes would be great here (one _minutes_ etc.)

tests/unit/utils.js
```diff
@@ -0,0 +1,60 @@
+ const assert = require('chai').assert;
+
+// create a global window so we can unittest the window.<x> functions
+if (!global.window)
+   global.window = new Object({});
+
+require('../../saltgui/static/scripts/utils');
+
```
oliverdunk: Is there a reason for two lines over one here?

tests/unit/parsecmdline.js
```diff
@@ -0,0 +1,371 @@
...
+    assert.equal(args[0], "NONE");
+    assert.equal(Object.keys(params).length, 0);
+
+    args = []; params = {};
+    result = window.parseCommandLine("NOne", args, params);
+    assert.isNull(result);
+    assert.equal(args.length, 1);
+    assert.equal(args[0], "NOne");
+    assert.equal(Object.keys(params).length, 0);
+
```
oliverdunk: I know I've left similar comments above, but is there a reason for the newlines after comments, or the multiple lines between code blocks? I just think there is room for some style improvements 😄

oliverdunk:
I'm running into an error when trying to use these tests locally:

```ERROR: error pulling image configuration: Get https://production.cloudflare.docker.com/registry-v2/docker/registry/v2/...: Service Unavailable```

The Cloudflare URL was quite long, so I've intentionally truncated it.

tests/unit/parsecmdline.js
erwindon: The unit tests in `tests/unit/parsecmdline.js` are grouped in sets. These comments are basically the group titles. 3 empty lines before that to visually separate from the previous group.

maerteijn:
@oliverdunk Seems like https:///production.cloudflare.docker.com was offline, as travis was able to download them. Could you try again?

I'll review your other comments later. The newlines come from the python style guide, which is unconsciously applied to javascript :-) Would be something to add to the linter btw.

tests/functional/login.js
maerteijn: Well, then the bash script should parse them, add them to the testing script which is implicitly invoked with yarn, then I need to parse the command again in javascript.

So I think WAY to much work (and hassle for me) for a small feature in the testrunner. But if you want it you can make a PR of course :smile:

tests/functional/login.js
maerteijn: Yes, I added it

package.json
maerteijn: Well, maybe not 😄 I removed it

tests/functional/login.js
maerteijn: No, I changed it.

runtests.sh
maerteijn: It's bash, not javascript :flushed: but I've added the newline.

saltgui/static/scripts/utils.js
maerteijn: Yes, I added it.

tests/functional/login.js
maerteijn: See https://github.com/segmentio/nightmare#end and the api documented there. See also the `beforeEach` where we queue the url.

tests/functional/login.js
maerteijn: I don't think it's a shame, it's actually really good. There are a many reasons why this is perfectly fine:

1) This test is incredibly easy to read as you immediately understand what is going on.
2) We expect this error message to be shown when you have the wrong credentials, so this is a logical thing to test for.
3) When the message is changed, or when the login sequence has changed (or when you made a mistake) this test will fail and forces you to investigate or to adjust the test.

tests/functional/login.js
maerteijn: Yes there should be, I added it.

tests/functional/login.js
maerteijn: Yes I added it.

tests/helpers/wait-for-docker.js
maerteijn: Well, the only way to do this with `node.js` (without adding dependencies) is with the shipped http library. (as `fetch` does not exist). And this module is really low level, so we have to make all the error handling ourselves (even getting the chunks of partial result data is something you have to do yourselves).

So know I rewrote it with the `requests` which is already a dependency of one of the other packages, which means we can drop `wait-on`.

tests/unit/utils.js
maerteijn: Yeah well, it would be fantastic but this function is now tested enough I think. PR's are welcome 😃

tests/unit/utils.js
maerteijn: No, so I adjusted it.

tests/unit/parsecmdline.js
maerteijn: I agree with oliver that there were 's a bit too much whitespaces. So I adjusted it a bit.

maerteijn:
So I reviewed all your comments and addressed most of the issues.

maerteijn:
And now I have to fix the build first :)

maerteijn:
Build is fixed.

maerteijn:
@oliverdunk Any chance to respond the following days? Would be nice to have your feedback and this pull request merged as @erwindon has a lot of work waiting for us 😄

oliverdunk:
Hey @maerteijn, absolutely. Will aim to do so tomorrow. If I'm blocking you and you're happy, feel free to merge without me 👍

PR #38, opened 2018-08-04
Added ASYNC and BATCH capabilities

erwindon:
Current functionality for the "Manual Run" box is to run a command synchronously. i.e. you wait until the command is complete. For some commands that may be a bit long. For different reasons, you may want to run the jobs piecewise on servers instead of all at once.
The PR add a drop-down box (initially only the menu triangle next to the "Run command" button) with the options Normal, Async and Batch. Once used, the selection remains visible. When Batch is selected, 2 additional drop-down menus are shown. That makes it possible to select the batch-size and the waiting time for starting new jobs. These 2 menu's only contain some fixed choices to keep things simple.
When the job is started as Async, a simple confirmation is shown listing the jobid (in a future PR the jobid becomes a link to the job page).
When the job is started as Batch, the system waits until all tasks are completed. No progress is show (as before). The results are combined and presented as if the job was started normally. This is a limitation of the API. (in a future PR this might become an incremental update with a progress indicator, when the salt message bus is eavesdropped).
These options are silently ignored when WHEEL or RUNNER functions are used.

### DIFF
--- a/saltgui/index.html
+++ b/saltgui/index.html
@@ -108,9 +108,9 @@
       <div class='run-command'>
         <div class='nearlyvisiblebutton' id='button_close_cmd'>&#x2715;</div>
         <h1>Manual Run</h1>
-        <input id='target' type='text' placeholder="Target"/>
-        <input id='command' type='text' placeholder="Command"/>
-        <input type='submit' value="Run command"/>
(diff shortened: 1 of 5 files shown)


erwindon:
Travis is failing after I accidentally pushed my branches into your repository; and then removed those again. Travis still thinks they exist...

Issue #40, opened 2018-08-14
How to treat JIDs in the run-command box?


erwindon:
SaltGUI has a popup panel where you can enter a salt commandline in which JSON datatypes are properly recognized. That also means that any job-id/JID that is naively entered becomes a number. The salt system is smart enough to translate the number back to a string; therefore no errors are shown. Unfortunately the job-ids are very long numbers and the last 4 digits of precision is lost.

e.g. job-id `20180814033130818988` is actually passed to the salt system as `20180814033130820000`. And that is a different (non-existing) job-id.

For an experienced user, the solution is simple: The job-ids are strings which should be quoted. Using `"20180814033130818988"` will prevent this problem.

There are a few actions possible:
* always treat sequences of 20 digits as a string; or
* show an error when a simple number has loss of precision; or
* leave the situation as-is.

My preference would be the first solution.
Any other solutions of suggestions?

maerteijn:
Hmm, interesting 😃  I would opt for solution one, as then we treat it as a identifier which it is, and not a integer value. 

So option 1 has my preference as well.

oliverdunk:
I'm a little late, but I agree. I can't see a situation where 20 or more consecutive digits are entered and they weren't looking to specify a job ID.

erwindon:
I did some more investigation here...
In Salt, it is the YAML parser in Python that decides that 20+ digits are a string. Thus automatically treating JIDs as strings. But that would not even matter, as there is also no loss of precision for numbers of that size.
It is the JS world of SaltGUI that loses precision at 17+ digits, which is consistent with the IEEE double format that JS uses.

erwindon:
Solved in #41

PR #41, opened 2018-08-14
Updated commandline parser to always treat strings that look like a JID as a string

erwindon:
As discussed in #40, the command parser is updated to always treat jobids as strings.
Since the jobid also looks like a number, it is otherwise treated as a number.
This prevents problems like `runners.jobs.lookup_jid 20180814033130818988` being executed as `runners.jobs.lookup_jid 20180814033130820000`.
Test with `test.echo 20180814033130818988`.
Note that the current version of SaltGUI barely uses JIDs; but that will change when `runners` commands are added.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .DS_STORE
 .gitignore
+yarn-error.log
 node_modules/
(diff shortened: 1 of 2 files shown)


maerteijn:
Could you add  a test for it now I merged the unittest branch? Then I'll merge it

erwindon:
unit tests are added

PR #42, opened 2018-08-19
Eslint and stylelint

maerteijn:
I introduced some css linting, see https://github.com/stylelint/stylelint-config-standard

Also I replaced jshint with eslint which is the defactor standard right now and easy to adjust the rules. I added some rules based on what makes sense + the feedback of @oliverdunk, see package.json in this pull request for details.

@erwindon Do you agree with these linting rules? Feel free to adjust in this pull request. When agreed, could you merge?

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,13 +10,13 @@
     "mocha": "^5.2.0",
     "nightmare": "^3.0.1",
     "stylelint": "^9.4.0",
-    "stylelint-config-recommended": "^2.1.0"
+    "stylelint-config-standard": "^18.2.0"
   },
   "jshintConfig": {
     "esversion": 6
   },
   "stylelint": {
-    "extends": "stylelint-config-recommended"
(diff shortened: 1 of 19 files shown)


erwindon:
The css checker works for me --> good
I see no problems with the added checks and all failure are already resolved --> good
I tested it with some re-introduced code --> good

One small problem left: `yarn.lock` should not be in git; thus it should be in `.gitignore`.
Note that I already added `yarn-error.log` to the `.gitignore` in branch `jobid`.

maerteijn:
Hi @erwindon 

I think yarn.lock SHOULD be in git. This ensures all dependencies of dependencies to be exactly the ones where we write software with.

erwindon:
If `yam.lock` is present on purpose, it's ok.
My remark was based only on fact that it is a generated file.
And I see too many cases in my other work where `.gitignore` is never updated and files are accidentally added.

PR #43, opened 2018-08-20
added sanity checks for startup of runtests.sh

erwindon:
Every time I run this script I get it wrong 2 times:
1) not using `bash`; and
2) not using `sudo`

It takes a few seconds and a bit of output before the script fails.
Therefore this update to let the script fail-fast.

### DIFF
--- a/runtests.sh
+++ b/runtests.sh
@@ -1,5 +1,11 @@
 #!/bin/bash
 
+# This script is a BASH script, therefore the interpreter must be BASH
+[ "$BASH" != "" ] || { echo "$0: must use bash"; exit 1; }
+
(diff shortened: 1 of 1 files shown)


erwindon:
The test for `bash` makes sense.
But it seems the test for `root` is not applicable to other systems than mine...

maerteijn:
I'm not sure why this is needed as:

```
#!/bin/bash
``` 
on top of the file says: 'execute this with bash' 

When users are not a member of the docker group they are not able to start any docker container at all. This is by the way mentioned in the docker documentation: https://docs.docker.com/install/linux/linux-postinstall/#manage-docker-as-a-non-root-user

PR #47, opened 2018-09-17
replace VAR with LET/CONST

erwindon:
I'm confusing myself often with the different use of `let` and `var`. Most of the new code that was contributed lately uses `let`. This PR changes all code to `let` and validates that code that no `var` still appear.

For viewers that want more details: see https://stackoverflow.com/questions/762011/whats-the-difference-between-using-let-and-var-to-declare-a-variable-in-jav

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,8 @@
       "no-multiple-empty-lines": ["error", { "max": 2, "maxEOF": 1 }],
       "no-useless-escape": 0,
       "no-constant-condition": ["error", { "checkLoops": false }],
-      "no-var": 2
+      "no-var": 2,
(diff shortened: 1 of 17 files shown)


maerteijn:
Nice!

You could also consider adding:
`"prefer-const": 2`

for vars that  are not mutated. But that I'll leave to you.

erwindon:
I created issue #48 for that. It is much more work since each assignment needs to be inspected individually. Let's stick to this content for this PR.

erwindon:
Ah, never mind.
`yarn jslint --fix` is great and does all the work...

maerteijn:
👍

PR #49, opened 2018-09-19
use only iterations instead of indexing

erwindon:
Some for-loops still use explicit indexing to access all list/array members, while the index itself is not used for anything else. In modern JS, this can be done in a more elegant way.
This PR converts all loops to more iteration-style loops.

e.g.
from:
`for(let i = 0; i < hostnames.length; i++) this._addMinion(list, hostnames[i]);`
to:
`for(const hostname of hostnames) this._addMinion(list, hostname);`


maerteijn:
Nice!

It also looks much better 😃 

(can you rebase?)

erwindon:
rebased!

PR #52, opened 2018-09-22
extension on documentation support: added external link support

erwindon:
In a recent PR, support for local documentation was added. In the conversion from the returned documentation to the presentation version, several markup features were applied. This still left a stray `_` character for external links, as well as a table of external links, each prefixed with the text `.. _`.
This PR recognized this situation and properly applies the link to the original text. The table of external links is now no longer visible.
This markup appears in e.g. `pkg.install`.

### DIFF
--- a/saltgui/static/scripts/output.js
+++ b/saltgui/static/scripts/output.js
@@ -242,6 +242,27 @@ class Output {
         // e.g. in "sys.doc state.template"
         out = out.replace(/>/g, "&gt;");
 
+        // external links
+        // e.g. in "sys.doc pkg.install"
+        while(out.includes(".. _")) {
+          // take only a line containing ".. _"
+          const reference = out.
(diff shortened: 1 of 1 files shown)


maerteijn:
Wouldn't it be much better to do this with a regular expression?

erwindon:
If we only wanted to remove the link table, that could have been done using just a regexp replacement. But we are also applying the links to the text fragments where they are used.

maerteijn:
Ok, pretty awesome feature!

Care to add a test for  it?

erwindon:
I've done a lot of visual testing. And a sample case is referenced in the comments for each transformation step. Can I leave it at that?

maerteijn:
I added a test for it https://github.com/maerteijn/SaltGUI/commit/194547dec5a083c554c12da4a5757ae73429cf50

PR #54, opened 2018-09-22
show progress of a long job

erwindon:
The jobs panel currently does not show any information about jobs that are still in progress. This PR adds the progress statistics for the jobs that are in progress. The number of running and the number of completed jobs is shown.
test with: start `salt '*' test.rand_sleep 300`. then refresh the SaltGUI screen.
note that this is static information, it is updated only on a screen refresh.

sample:
> afbeelding

### DIFF
--- a/saltgui/static/scripts/routes/page.js
+++ b/saltgui/static/scripts/routes/page.js
@@ -155,13 +155,7 @@ class PageRoute extends Route {
       const job = jobs[k];
 
       // start with same text as for _addJob
-      let targetText = job.Target;
-      if(job["Target-type"] !== "glob" && job["Target-type"] !== "list") {
-        // note that due to bug in 2018.3, all finished jobs
-        // will be shown as if of type 'list'
-        // therefore we suppress that one
-        targetText = job["Target-type"] + " " + targetText;
-      }
+      let targetText = window.makeTargetText(job["Target-type"], job.Target);
 
       // then add the operational statistics
       if(job.Running.length > 0)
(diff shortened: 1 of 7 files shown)


saltgui/static/scripts/routes/page.js
```diff
@@ -146,9 +148,35 @@ class PageRoute extends Route {
...
+    {
+      const job = jobs[k];
+
+      // start with same text as for _addJob
+      let targetText = job.Target;
+      if(job["Target-type"] !== "glob" && job["Target-type"] !== "list") {
+        // note that due to bug in 2018.3, all finished jobs
+        // will be shown as if of type 'list'
+        // therefore we suppress that one
+        targetText = job["Target-type"] + " " + targetText;
```
maerteijn: Wouldn't it be an idea to create a function for this? Now I see this for the third time in de codebase when I'm not mistaking

erwindon: It is in there 2 'only' times. Once for the overview; and once for the detailed view. Still reason enough to create a function for it... please wait...

erwindon:
PR is updated with the suggested code change.
Also rebased.
PR is valid again...

maerteijn:
Looks good, could you add a unittest for  makeTargetText ?

erwindon:
ok, hang on...

erwindon:
added unit tests for all targettypes.
rebased to hide a nasty error.
added screen fragment to initial PR text.
the PR is valid again

saltgui/static/scripts/routes/page.js
maerteijn: So here you should also use `window.makeTargetText` right?

erwindon:
certainly!

PR #56, opened 2018-09-25
Added grains page

erwindon:
This PR adds a simple page with information about the grains.
No operations are defined on the grains yet.
A summary of the additional grains is shown on the minion list.
The list of grains can be shown using a (now) traditional dropdown menu.
One can either show only the custom grains; or all grains.
An alternative for JSON.stringify library is introduced that shows dictionaries with sorted keys.
Some helper functions for the minions and keys pages were moved to the corresponding JS file.

### DIFF
--- a/saltgui/index.html
+++ b/saltgui/index.html
@@ -6,6 +6,7 @@
     <link rel='stylesheet' type='text/css' href='static/stylesheets/login.css' />
     <link rel='stylesheet' type='text/css' href='static/stylesheets/page.css' />
     <link rel='stylesheet' type='text/css' href='static/stylesheets/job.css' />
+    <link rel='stylesheet' type='text/css' href='static/stylesheets/grains.css' />
     <link rel='stylesheet' type='text/css' href='static/stylesheets/dropdown.css' />
     <link rel='shortcut icon' href='static/images/icon.png' />
   </head>
@@ -16,6 +17,7 @@
       <div class='menubar'>
         <div class='menu_item' id='button_minions'>minions</div>
         <div class='menu_item' id='button_keys'>keys</div>
+        <div class='menu_item' id='button_grains'>grains</div>
         <div class='menu_item' id='button_logout'>logout</div>
       </div>
       <div class='fab' id='button_manualrun'>&gt;_</div>
@@ -91,6 +93,23 @@
         </div>
       </div>
 
+      <div class='route' id='page_grains'>
+        <div class='dashboard'>
+          <div class='panel minion-list'>
+            <h1>Grains</h1>
+            <ul id="minions" class='minions'></ul>
+          </div>
+          <div class='panel job-list'>
+            <h1>Recent Jobs</h1>
+            <ul class='jobs'></ul>
+          </div>
+          <div class='panel grains-list'>
+            <h1 id="grains_title">Grains</h1>
+            <ul id="grains_list" class='grains'></ul>
+          </div>
+        </div>
+      </div>
+
       <div class='route' id='page_job'>
         <div class='panel job-info'>
           <div class='nearlyvisiblebutton' id='button_close_job'>&#x2715;</div>
@@ -114,6 +133,7 @@
     </div>
 
     <script src='static/scripts/utils.js'></script>
+    <script src='static/scripts/stringify.js'></script>
     <script src='static/scripts/parsecmdline.js'></script>
     <script src='static/scripts/output.js'></script>
     <script src='static/scripts/documentation.js'></script>
@@ -125,6 +145,7 @@
     <script src='static/scripts/routes/job.js'></script>
     <script src='static/scripts/routes/minions.js'></script>
     <script src='static/scripts/routes/keys.js'></script>
+    <script src='static/scripts/routes/grains.js'></script>
     <script src='static/scripts/dropdown.js'></script>
     <script src='static/scripts/router.js'></script>
   </body>
--- a/saltgui/static/scripts/routes/grains.js
+++ b/saltgui/static/scripts/routes/grains.js
@@ -1,0 +1,97 @@
+class GrainsRoute extends PageRoute {
+
+  constructor(router) {
+    super("^[\/]grains$", "Grains", "#page_grains", "#button_grains");
+    this.router = router;
+    this.keysLoaded = false;
+    this.jobsLoaded = false;
+
+    this._showGrains = this._showGrains.bind(this);
+    this._updateJobs = this._updateJobs.bind(this);
+    this._updateKeys = this._updateKeys.bind(this);
+    this._updateMinion = this._updateMinion.bind(this);
+  }
+
+  onShow() {
+    const minions = this;
+
+    const grainsContainer = document.querySelector("#page_grains .grains-list");
+    grainsContainer.style.display = "none";
+
+    return new Promise(function(resolve, reject) {
+      minions.resolvePromise = resolve;
+      if(minions.keysLoaded && minions.jobsLoaded) resolve();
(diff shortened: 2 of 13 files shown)


saltgui/static/scripts/routes/grains.js
```diff
@@ -0,0 +1,242 @@
...
+      "pkgsrcversion",
+      "product",
+      "roles",
+      "systemfirmware",
+      "systemfirmwaredate",
+      "virtual_subtype",
+      "zpool",
+    ];
+  }
+
```
maerteijn: Isn't there a api call to retrieve a list of available grains?

erwindon: Searched and found nothing about that in the API. Then did a source code check to find them all, including those that are only returned under certain circumstances. A bridge too far?

maerteijn: Don't think it's a bridge too far as it it's a really handy feature. I dislike the fact that we need to hardcode a grain list while this is depending on the salt version, and that grains can be custom as well.

```grains.ls``` will show the grains right? isn't there something we can do?

Comment (erwindon) [reply saltgui/static/scripts/routes/grains.js]: `grains.ls` shows all grains whether predefined or custom

**Comment by erwindon** [comment]
The sorting is for the minor sorting only. i.e. the grains themselves can be sorted using their name. The sorted representation was needed for grains like `dns`, `gpus`, `locale_info` and only a few others that have a dict value inside. let's take that functionality out for now. I think it is not that difficult to write my own version.

**Comment by erwindon** [comment]
rebased

File: saltgui/static/scripts/routes/grains.js (hunk 1 above)
Comment (maerteijn) [reply saltgui/static/scripts/routes/grains.js]: `grains.ls shows all grains whether predefined or custom.`

That's a plus right? :) What would be the problem with  that?

Comment (erwindon) [reply saltgui/static/scripts/routes/grains.js]: I missed the fact that there was confusion at the start of this thread.
The list exists *only* to distinguish the built-in grains from the custom grains. If that distinction is not worth the trouble, then it can go. My remark *a bridge too far* was only about using the list to distinguish the built-in grains from the custom grains.

**Comment by erwindon** [comment]
bug detected in the documentation detector, let's fix that first...

**Comment by erwindon** [comment]
changes made in the latest commit, PR is valid again

File: saltgui/static/scripts/routes/grains.js [hunk 2]
```diff
@@ -0,0 +1,99 @@
...
+      const name = Route._createDiv("grain_name", k);
+      grain.appendChild(name);
+
+      const grain_value = Stringify.format(minion[k]);
+      const value = Route._createDiv("grain_value", grain_value);
+      grain.appendChild(value);
+
+      container.appendChild(grain);
+      cnt++;
```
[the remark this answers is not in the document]
erwindon: That was a remnant of the era where you could chose between *all* grains and only the *custom* grains. With 0 custom grains an additional text was then shown. So, yes, that should have been removed. It is now removed.

saltgui/static/scripts/stringify.js
```diff
@@ -0,0 +1,29 @@
...
+    }
+    else {
+      // use a multi-line format with indentation
+      txt = JSON.stringify(a, null, "  ");
+    }
+
+    return txt;
+  }
+
+}
```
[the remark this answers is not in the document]
erwindon: Unit tests will be prepared soon, please wait...

erwindon:
added unit tests for stringify, fixed 2 small issues, and rebased.
PR is valid again.

saltgui/static/scripts/stringify.js
```diff
@@ -0,0 +1,32 @@
...
+
+// for unit tests
+if(typeof module !== "undefined") module.exports = Stringify;
```
maerteijn: I've also noticed this in the documentation class. Isn't this something we can avoid?

saltgui/index.html
```diff
@@ -114,6 +133,7 @@ <h1>Manual Run</h1>
     </div>
 
     <script src='static/scripts/utils.js'></script>
+    <script src='static/scripts/stringify.js'></script>
```
maerteijn: Would it  be an idea to add this function to the `output.js`file ?

saltgui/static/scripts/stringify.js
erwindon: As the comment says, you need this for the unit tests.
This is basically the equivalent of the `<script ... >` statement from `index.html`.

saltgui/index.html
erwindon: yes. but the functionality in `stringify.js` is not complete yet. It currently does not sort the dictionary keys, and it does not properly format the empty arrays and dictionaries that are embedded. That is on my backlog. When that functionality is complete I'll integrate the two when that is still reasonable. (i'll create a story for that)

erwindon:
I made this a separate page, because:
* some operations will be placed on it later to manipulate the grains. that is not something you want to do inside the regular output panel
* since editing is often per minion, I wanted the minion list on the screen. That is also more consistent with the other pages.
the job list was hidden because it is not the main topic of that page. but until a minion is selected it fills the space nicely

erwindon:
needs refactor due to making the object formatter a separate PR...

maerteijn:
"the job list was hidden because it is not the main topic of that page. but until a minion is selected it fills the space nicely"

Not sure about that. There is now not a logical way to show the jobs. It feels not right to do it like it is now implemented.

Based on your input I then would say: Move the jobs to a own page. Then you can show the grains and pillar information on the right side as you implemented now. I still think you should not make a separate "grains" and "pillar" page, but add it to the existing minion dropdown menu as suggested.

erwindon:
> Then there is not a logical way to show the jobs and feels not well thought out.
this happens only on the "grains" page (and "pillars" pages, and "schedules" page).
the jobs info remains as-is on the main page (and even on the "keys" page).
I'll update the page flow so that the minions on the main page get a menu item "grains" that navigates directly to a "grains-details" page.
The old "grains" page will be updated to still show the overview, but on selection also navigate to the corresponding "grains-details".
For me, the overview of number of grains is essential. without it, I am forced to go through all my minions one-by-one, This page is why I joined this project.

maerteijn:
@erwindon 

> Then there is not a logical way to show the jobs and feels not well thought out.

Having a OWN menu item "jobs" so that we can extend the minions page with functionality which is only minion related sounds pretty well thought out to me. Showing jobs "to fill out space nicely" sounds not well thought out to me. Further all, I don't like the additions of the grains and pillar menus as I don't see those a main topics.

But let's stop wasting each others time. This application is now pretty important to you hence the amount of time on you spend on it. I'm maintaining it out of courtesy, for fun and for learning things.

But now I spend way too much time on reviewing your pull requests, on refactoring, on discussions how things should work and I have the idea I can't keep up with all the things you already have planned and done. Your coding style also differs a lot, which is an opinionated thing, but  this takes me also a lot of time. I'm not using the project actively and this will not change in the near future, so:

I decided to stop development on it for a while and I will transfer this project to you. Please take good care of it and I wish you good luck with the further development.

maerteijn:
(PS could you push all you branches to my repo and then delete your fork? Then I can transfer the ownership of this repository to you)

erwindon:
(all branches pushed, my fork is deleted)

maerteijn:
Here you go, thanks for the kind words and have fun with developing SaltGUI!

erwindon:
thx!
can you please go through the issues/PRs and re-assign all those you will not be working on now?

maerteijn:
I will

erwindon:
github is confused about the branch, probably due to the change-owner of the repository

PR #57, opened 2018-09-26
fixed array output

erwindon:
When issuing the `grains.ls` command, its output is mistaken for documentation output.
The documentation detector (in `output.js`) did not test for an object to actually being an array.
Now it does, and a unit test was added to prove it.

### DIFF
--- a/tests/unit/output.js
+++ b/tests/unit/output.js
@@ -28,6 +28,11 @@ describe('Unittests for output.js', function() {
     result = Output.isDocumentationOutput(outputData, "keyword");
     assert.isFalse(result);
 
+    // documentation is not text in a dict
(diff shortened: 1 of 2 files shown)


tests/unit/output.js
```diff
@@ -28,6 +28,11 @@ describe('Unittests for output.js', function() {
     result = Output.isDocumentationOutput(outputData, "keyword");
     assert.isFalse(result);
 
+    // documentation is not text in a dict
```
maerteijn: `in a dict` should be `in an array` right?

erwindon: it is correct. but still confusing. so updated all comments.

PR #64, opened 2018-10-03
Cleanup API class, catch promise rejections where needed, documented permissions settings

maerteijn:
A lot of changes here, but I think for the better to make the code more maintainable and easier to adjust.

Including:
- Permission documentation
- Cleanup of (some) unnecessary `Promises` in the API class 
- Separated the commandbox in a separate class to make the API class, well, API only
- Let `isAuthenticated` do the checks if we need to login or not and let the  pages react on it
- Fixed the `Property 'handleEvent' is not callable` error (finally!)
- Fixed some issues with the hamburger menu
- Simplified some stuff here and there

I think some more cleanup is needed  and error handling could be a bit better but let's do this first.


saltgui/static/scripts/api.js
```diff
@@ -116,24 +119,24 @@ class API {
   }
 
   isAuthenticated() {
-    return window.sessionStorage.getItem("token") !== null;
+    // As there is not an API call to check if you are authenticated we call
+    // the stats call to see if we can access that
+    return this._callMethod("GET", "/stats", {})
```
erwindon: normally, a function like `_fetch` would just react on no sessionkey or when 301 returned.
it should be possible to eliminate `isAuthenticated` completely.

saltgui/index.html
```diff
@@ -108,7 +108,6 @@ <h1>Manual Run</h1>
         <div><input id='target' type='text' placeholder="Target"/></div>
         <div id="cmdbox"><input id='command' type='text' placeholder="Command"/></div>
         <input type='submit' value="Run command"/>
-        <div id="cmd_error"></div>
```
erwindon: Nice solution. I was in doubt about that when I added the error-div.

saltgui/static/scripts/router.js
```diff
@@ -92,5 +92,5 @@ class Router {
 
 }
 
-window.addEventListener('load', new Router());
+window.addEventListener('load', () => new Router());
```
erwindon: I spend at least an hour on that too. decided it was not caused by SaltGui. guess I was wrong...

README.MD
```diff
@@ -19,17 +18,22 @@ A new open source web interface for managing a SaltStack server. Built using van
...
   pam:
-    desiredSaltGuiUsername:
-      - .*
-      - '@runner'
-      - '@wheel'
+    saltuser:
+      - grains.items
+      - sys.doc
+      - state.apply
+
```
erwindon: this is a straitjacket.
nothing wrong with an explanation (in `PERMISSIONS.md`) about several restricted scenarios.
but the original configuration-set matched the commandline version of `salt`; and that is now gone.
so, whoever really wants to use SaltGUI must do investigation and more configuration to get things done.

maerteijn: Yes, maybe you are right: I should leave the default configuration as it was. I will update that.

maerteijn:
@erwindon I updated the documentation with your comments in mind. You can merge this pull request if you think it's ok

Issue #65, opened 2018-10-04
improve object print


erwindon:
The current object printer uses the default formatter, which:
* does not order the keys of a dictionary
* has slightly unreadable presentation for empty array `[]` and empty dictionary `{}`
An updated formatter should order the keys and place a little bit of space between the braces.

e.g. now:
![afbeelding](https://user-images.githubusercontent.com/3663742/46494413-11996700-c813-11e8-9895-b669fd70ac8c.png)

the code should be update to sort the keys of a dictionary, and to put some spacing between the bracket pairs.

PR #66, opened 2018-10-06
improve object print

erwindon:
As described in issue #65, the current object printer uses the default formatter, which:
* does not order the keys of a dictionary
* has slightly unreadable presentation for empty array `[]` and empty dictionary `{}`

This PR enhanced both aspects.
The new formatter is applied to the output from the command box and the job viewer.
The new formatter is not applied to the objects passed with the API.
As suggested elsewhere, the implementation is completely in `output.js`.

e.g. a part of the new `grains.items` output:
```
host01: {
  "SSDs": [ ],
  "biosreleasedate": "06/02/2017",
  "biosversion": "090007",
...
```

The PR also fixes an issue where an array was treated as an object. that is technically true, but not what we intended at that point.

This PR also fixes an issue with the use of `trimEnd`. The equivalent function `trimRight` has better availability and is now used instead.

### DIFF
--- a/saltgui/static/scripts/output.js
+++ b/saltgui/static/scripts/output.js
@@ -23,7 +23,11 @@ class Output {
 
   // format a JSON object
   // based on an initial indentation and an indentation increment
-  static _formatJSON0(value, indentStr, indentCnt) {
+  // default initial indentation is ""
+  static formatJSON(value, indentStr="") {
+
+    // indent each level with 2 spaces
+    const indentStep = "  ";
 
     if(typeof value !== "object") {
       // a simple type
@@ -37,8 +41,8 @@ class Output {
       let str = "[";
       let separator = "";
       for(const elem of value) {
-        str += separator + "\n" + indentStr + indentCnt +
-          Output._formatJSON0(elem, indentStr + indentCnt, indentCnt);
+        str += separator + "\n" + indentStr + indentStep +
+          Output.formatJSON(elem, indentStr + indentStep);
         separator = ",";
       }
       if(value.length === 0) {
@@ -57,8 +61,8 @@ class Output {
     let str = "{";
     let separator = "";
     for (const [key, val] of Object.entries(value).sort()) {
-      str += separator + "\n" + indentStr + indentCnt + "\"" + key + "\": " +
-        Output._formatJSON0(val, indentStr + indentCnt, indentCnt);
+      str += separator + "\n" + indentStr + indentStep + "\"" + key + "\": " +
+        Output.formatJSON(val, indentStr + indentStep);
       separator = ",";
     }
     if(Object.keys(value).length === 0) {
@@ -69,13 +73,6 @@ class Output {
     }
     str += "}";
     return str;
-  }
-
-  // format a JSON object
-  static formatJSON(a) {
-    // initial indentation is ""
-    // indent each level with 2 spaces
-    return Output._formatJSON0(a, "", "  ");
   }
 
   // Re-organize the output to let it appear as if the output comes


saltgui/static/scripts/output.js
```diff
@@ -110,6 +167,12 @@ class Output {
         return false;
       }
 
+      // arrays are also objects,
+      // but not what we are looking for
+      if(Array.isArray(output)) {
+        return false;
+      }
+
```
maerteijn: This code is now duplicated: line 164 and line 170

saltgui/static/scripts/output.js
```diff
@@ -21,6 +21,63 @@
 
 class Output {
 
+  // format a JSON object
+  // based on an initial indentation and an indentation increment
+  static _formatJSON0(value, indentStr, indentCnt) {
```
maerteijn: Can we just have a `static formatJSON(json, indent=4)` function here instead of a public and private one?

Difference between initial indentation and other indentation would be something I avoid. The indentation param should be a number and not a string.

saltgui/static/scripts/output.js
```diff
@@ -21,6 +21,63 @@
...
+    }
+    str += "}";
+    return str;
+  }
+
+  // format a JSON object
+  static formatJSON(a) {
+    // initial indentation is ""
+    // indent each level with 2 spaces
+    return Output._formatJSON0(a, "", "  ");
```
maerteijn: See first comment about function name and arguments

saltgui/static/scripts/output.js
```diff
@@ -21,6 +21,63 @@
...
 class Output {
 
+  // format a JSON object
+  // based on an initial indentation and an indentation increment
+  static _formatJSON0(value, indentStr, indentCnt) {
+
+    if(typeof value !== "object") {
+      // a simple type
+      // leave that to the builtin function
+      return JSON.stringify(value);
```
maerteijn: Add the indentation here from the `indent` parameter

saltgui/static/scripts/output.js
```diff
@@ -21,6 +21,63 @@
...
+      }
+      if(value.length === 0) {
+        // show the brackets for an empty array a bit wider apart
+        str += " ";
+      } else {
+        str += "\n" + indentStr;
+      }
+      str += "]";
+      return str;
+    }
```
maerteijn: I don't see any sorting applied here so why not just
```json
JSON.stringify([1,2,3,4,5], null, indent)
"[
    1,
    2,
    3,
    4,
    5
]"
```

saltgui/static/scripts/output.js
```diff
@@ -21,6 +21,63 @@
...
+    // regular object
+    // put each name+value on its own line
+    const keys = Object.keys(value);
+    let str = "{";
+    let separator = "";
+    for (const [key, val] of Object.entries(value).sort()) {
+      str += separator + "\n" + indentStr + indentCnt + "\"" + key + "\": " +
+        Output._formatJSON0(val, indentStr + indentCnt, indentCnt);
+      separator = ",";
+    }
```
maerteijn: Here I see the sorting indeed so that would make sense. But wouldn't it be better if you just sort it and then feed it to `formatJSON` which uses `JSON.stringify` ?

saltgui/static/scripts/output.js
erwindon: because the one of the values of array may be an object which needs sorting

saltgui/static/scripts/output.js
erwindon: Of course. I'm not used to languages with actual default values for their parameters. Hence the strict separation.

saltgui/static/scripts/output.js
erwindon: That because I do not trust JS. Better do it twice...

erwindon: just kidding, it was duplicated due to cherry-pick. then after the merge we had both. cleaned up.

saltgui/static/scripts/output.js
erwindon: I kept the the indentation parameters as string. the are used in multiple places and I don't want to use `' '.repeat(indentCnt)` each time.

saltgui/static/scripts/output.js
erwindon: (but that was changed due to an other remark)

saltgui/static/scripts/output.js
erwindon: the indentation is only added for the content of arrays/objects.
otherwise the value `123` in `{"a":123}` would also be indented.
strings, numbers, the starting `[` and the starting `{` are not indented.

erwindon:
back to you...

maerteijn:
I'd still would say: use an integer for the indentation amount instead of a string, but it's your call now.

erwindon:
> I'd still would say: use an integer for the indentation amount instead of a string, [...]
done, thx.

PR #70, opened 2018-10-07
simplified reload-on-change

erwindon:
When using the commandbox and running a command that may have altered the screen content, a screen refresh is done when the commandbox closes. The current implementation is simple and effective, but not easily extendible.
The current test basically tested for commands starting with `wheel.key.` using a simple `if` statement.
This PR replaces that test with a test against an explicit list. The list is easily extendible when new commands should be tested for.

Note that this mechanism remains a bit primitive:
* it works only on the last command executed: a `wheel.key.delete` followed by a `test.ping` in the same commandbox does not cause the screen refresh
* it works without considering the current page. but typically a command is only easily available from a page where it is relevant

### DIFF
--- a/saltgui/static/scripts/commandbox.js
+++ b/saltgui/static/scripts/commandbox.js
@@ -85,10 +85,15 @@ class CommandBox {
 
     // test whether the command may have caused an update to the list
     // the user may have altered the text after running the command, just ignore that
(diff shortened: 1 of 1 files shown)


erwindon:
ok, will fix that now

Issue #11, opened 2018-05-15
Feature request: Job Template or Favourite Jobs


lostsnow:
I'm using [SaltPad](https://github.com/Lothiraldan/saltpad/), but this repository is not currently maintained. It has a feature called "Job Template" that is very useful, that used to execute some common commands.

Can you add similar feature to execute some frequently used commands or jobs?

erwindon:
should the saved information contain both the target and command information? or just the command?
saltgui is a client-side application. is saving in the browser's datastore sufficient? that fixes the information to one specific browser and it may be readable for others.

erwindon:
I just studied the info for SaltPad. The templates are in the main configuration file. See https://github.com/Lothiraldan/saltpad/blob/master/settings.json.sample. It contains named tuples containing targettype, targetpattern, command and arguments. It should not be that much work to read a similar configuration file (whether in json or yaml) and apply it to the fields of the "run command" dialog box.
@lostsnow : you did not respond to the questions from May 22nd. Is the request still valid?

lostsnow:
Yes, it is not very difficult work but very useful for frequently used commands.

maerteijn:
@erwindon You mean that you can define the templates in the saltmaster master file configuration, and that you can use them in saltgui?  Pretty neat

erwindon:
yes.
Thanks for SaltPad for the original idea.
Thanks for @lostsnow for the request.

PR #82, opened 2018-11-20
added support for command templates

erwindon:
This PR adds support for templates for the command-box.
As requested in #11.
Templates are maintained in the `master` file.
Values from the selected template are simply copied into the on-screen fields.

### DIFF
--- a/README.MD
+++ b/README.MD
@@ -60,6 +60,26 @@ SaltGUI supports the following authentication methods supported by salt:
 
 See the [EAUTH documentation](https://docs.saltstack.com/en/latest/topics/eauth/index.html) and the [Salt auth source code](https://github.com/saltstack/salt/tree/2018.3/salt/auth) for more information.
 
+## Templates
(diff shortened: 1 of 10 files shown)


erwindon:
@lostsnow it took a while to honour your request. Some internal plumbing enhancements and some of our own feature requests came first. But now the support for simple templates is there. Do you care to do a functional and/or technical review?

lostsnow:
How about adding a vertical scroll bar when there are too many templates? just add `max-heiget` and `overflow-y` to css.

erwindon:
Seems you are making good use of the templates menu.
I already had scrolling menus in another experimental branch, with the same code as you suggest.
So, just a few minutes please...

erwindon:
The scrolling menu is now there.
Can you please leave a note also when you think no further changes are needed in this area?

lostsnow:
Great, no further changes are needed.

PR #83, opened 2018-11-25
add nodegroup to target type

lostsnow:
Signed-off-by: lostsnow <lostsnow@gmail.com>

### DIFF
--- a/README.MD
+++ b/README.MD
@@ -64,7 +64,7 @@ See the [EAUTH documentation](https://docs.saltstack.com/en/latest/topics/eauth/
 SaltGUI supports command templates for easier command entry into the command-box.
 The menu item for that becomes visible there when you define one or more templates
 in salt master configuration file `/etc/salt/master`.
-The field `targettype` supports the values `glob`, `list` and `compound`.
+The field `targettype` supports the values `glob`, `list`, `compound` and `nodegroup`.
 Entries will be sorted in the GUI based on their key.
 You can leave out any detail field.
 e.g.:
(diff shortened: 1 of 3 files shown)


erwindon:
Never mind the errors, these are not caused by your code.

erwindon:
You code looks good. But it won't work with the templates yet. I'll add the missing part in 
https://github.com/erwindon/SaltGUI/blob/79fe1ca0ea306ea8779ce9e88f32dcc725829b82/saltgui/static/scripts/commandbox.js#L66

erwindon:
We want to keep SaltGUI as simple as possible. That is why the templates-menu is not visible when there are no templates. The targettype `nodegroup` should not be visible when there are no nodegroups.
In `minions.js` we load the templates into memory. The same should be done there for nodegroups.
In the constructor in `targettype.js` something similar to lines 22-24 from `commandbox.js` must be done.

erwindon:
Branch `templates` has been merged, no need to rebase to there.

erwindon:
@lostsnow can you please rebase your feature branch?
rebasing is a much better alternative to merging a master branch into a feature branch...

erwindon:
replaced with #85

erwindon:
@lostsnow 
I'm thinking about improving on the nodegroups a bit further. Please let me know your opinion on the following...
1) Currently the target-type is automatically selected when you type part of a List (by typing a comma) or a Compound (by typing @, blank or parentheses). I think it would be useful to auto-select the target-type NodeGroup also. So now whenever you type a `#` as first character, that happens and anything after the `#` is used as the nodegroup name.
2) When you are using a non-existing nodegroup, an error-message is produced. Would it be useful to list the names of the known nodegroups with that message? maybe only until a maximum amount of groups.
What is the order of number of nodegroups in your case?

lostsnow:
How about show a autocomplete select box when typing start with `#`? (auto-select the target-type NodeGroup also), just limit the number of choices for large number of nodegroups.

PR #85, opened 2018-12-01
Nodegroup

erwindon:
Based on #83 and the first implementation by @lostsnow, the commandbox now supports nodegroups.
The menuitem for that only appears when there are nodegroups defined.
A custom error message is shown when the node group is unknown, which is better than the "Internal Server Error" that the SaltAPI would return.


lostsnow:
It works ok, thanks

Issue #86, opened 2018-12-05
simplify nodegroup-name entry


erwindon:
from @lostsnow:
How about show a autocomplete select box when typing start with #? (auto-select the target-type NodeGroup also), just limit the number of choices for large number of nodegroups.

erwindon:
@lostsnow what is the rough number of nodegroups that you use?

erwindon:
@lostsnow: I added a simple dropdown box to the target field. See #89. Can you also do a little test with that branch?

erwindon:
thx @lostsnow for ideas and testing!

PR #89, opened 2018-12-09
added autocomplete for minion or nodegroup entry

erwindon:
As requested in #86 
Uses the html native feature `<datalist>`.
The list holds the minion names and the node-group names. The node-group names are prefixed with `#` for easy entry.
Note that both the minion list and the node-group list are updated asynchronously. So entries in the list may be available only a few seconds after start-up.


lostsnow:
It looks good to me, thanks.

PR #114, opened 2018-12-30
Sortable

erwindon:
Updated gui to use tables instead of div collections. That should make sorting much easier. Layout has automatically improved nicely.
Some rework was needed on the Keys page due to the use of sub-headers.
Some rework was needed to keep the dropdown menu's working properly when inside a table.
next step is to add the new column for minion salt version (done)
next step is to add table header (done)
next step is to add the actual sorting (done)


dawidmalina:
Tested. Works perfectly 👍

erwindon:
thx!
just one more thing before I can merge: the `Keys` pages becomes messy when sorting because the sorting algorithm does not know about the sub-headers.

Issue #119, opened 2018-12-31
Hardcode docker container hostname


dawidmalina:
It would be nice to have fixed (hardcoded) hostname which (by default) will be used as a minion_id. Hostname will match name in docker-compose.yaml.

For example instead:
```
6884dd9bd3c9
```

we will have:
```
saltminion-ubuntu
```

<img width="925" alt="zrzut ekranu 2018-12-31 o 15 56 50" src="https://user-images.githubusercontent.com/5910851/50562576-28b5fb00-0d15-11e9-881e-6f8c85b3cdf9.png">

erwindon:
I see you have PR #120 for this...

erwindon:
merged

PR #120, opened 2018-12-31
Hardcode docker container hostname #119

dawidmalina:


erwindon:
tested, looks good in browser. not expecting any side-effects since the previous hostnames were random anyway.

Issue #118, opened 2018-12-31
copying ip numbers does not work


erwindon:
message "copied!" is shown, but clipboard only contains an HTML-node which cannot be pasted as text.
probably due to migration to tables?
that is still in development, so this may not be more than a reminder for that...

erwindon:
See PR #122

erwindon:
merged

PR #122, opened 2019-01-01
removed css that prevented copy2clip to work

erwindon:
See #118 
Solved by removing the `user-select: none` clauses from CSS.
@dawidmalina can you verify that this now works properly?


dawidmalina:
Tested. Works perfectly 👍

Issue #107, opened 2018-12-27
UX sync_grains from grains information page


dawidmalina:
I think that it would be nice improvement if we could `sync_grains` for particular minion from the grains page by extending existing menu

<img width="932" alt="zrzut ekranu 2018-12-27 o 16 37 42" src="https://user-images.githubusercontent.com/5910851/50485463-c1870680-09f5-11e9-8235-a53dae27a1bc.png">

dawidmalina:
The best best approach would be to extend menu on the grains page details with new menu item which will allows us to call `saltutil.sync_grains`. We usually found this helpful (if you apply grains from master side) to force sync up of the grains.
<img width="930" alt="zrzut ekranu 2018-12-31 o 02 54 58" src="https://user-images.githubusercontent.com/5910851/50553475-cf63b280-0ca7-11e9-80eb-b7d5b224c45b.png">

erwindon:
This is for a specific minion.
Is this useful for all minions also?

erwindon:
I had the following choices:
1) On `/pillars` add `saltutil.sync_grains` for `*` (top of page)
2) On `/pillars` add `saltutil.sync_grains` for specific minion (on minion row)
3) On `/pillarsminion` add `saltutil.sync_grains` for that minion (top of page)
So only (3) is actually desired, correct?

That command behaves properly when I try this.

erwindon:
Should the page `/pillarsminion` be refreshed after the command was finished (and the run-box closed)?

dawidmalina:
Yes, we should focus only on 3rd option (`/grainsminion` page) and exactly :) page should be refreshed and run-box can be closed.

erwindon:
see #126

erwindon:
merged

Issue #106, opened 2018-12-27
UX sync_pillar from pillar information page


dawidmalina:
I think that it would be nice improvement if we could `sync_pillar` for particular minion from the pillar page by extending existing menu

<img width="922" alt="zrzut ekranu 2018-12-27 o 16 35 54" src="https://user-images.githubusercontent.com/5910851/50485416-84bb0f80-09f5-11e9-8e04-6476be2e743d.png">

erwindon:
Can you please describe a use-case?

dawidmalina:
The best best approach would be to extend menu on the pillar page details with menu item which will allows us to call `saltutil.sync_pillar`. We usually found this helpful (if you apply pillar from external side aka external pillars) to force sync up of the pillars.

erwindon:
This is for a specific minion.
Is this useful for all minions also?

dawidmalina:
Main goal is to have this per minion which can be refreshed from pillars details page. Similar to what i described in grains ticket. Use caae for this is to force sync if pillars are not yet as expected.

erwindon:
I had the following choices:
1) On `/pillars` add `saltutil.sync_pillar` for `*` (top of page)
2) On `/pillars` add `saltutil.sync_pillar` for specific minion (on minion row)
3) On `/pillarsminion` add `saltutil.sync_pillar` for that minion (top of page)
So only (3) is actually desired, correct?

The command shows the following error returned form the minion when I try this:
`ERROR: Pillar modules can only be synced to masterless minions`
(help!)

erwindon:
Should the page `/pillarsminion` be refreshed after the command was finished (and the run-box closed)?

dawidmalina:
Yes, we should focus only on 3 option (`/pillarsminion` page) and in this case we should use command: `saltutil.refresh_pillar` and exactly :) page should be refreshed and run-box can be closed.

erwindon:
see #126

erwindon:
merged

PR #126, opened 2019-01-03
added refresh_pillar+sync_grains commands

erwindon:
see #106 and #107


dawidmalina:
Do you think we could switch from `saltutil.sync_grains` to `saltutil.refresh_grains` to same approach (with naming and method)?

dawidmalina:
Yes. And I just wondering would it be hard to have auto closing run command box window whenever result of the command is successful?

erwindon:
That is a completely independent change. Feel free to create a separate issues for it.

dawidmalina:
LGTM

erwindon:
merged

Issue #99, opened 2018-12-26
UX grains information


dawidmalina:
I think that user experience with displaying grains information could be even better if we could format output instead printing raw array, map
<img width="917" alt="zrzut ekranu 2018-12-26 o 20 55 41" src="https://user-images.githubusercontent.com/5910851/50455850-f0976c80-0950-11e9-91c7-c025f3d5f579.png">

dawidmalina:
For example instead:
```
[
  "nbd3",
  "nbd15",
  "nbd1",
  "nbd13",
  "nbd8",
  "nbd11",
  "nbd6",
  "nbd4",
  "nbd2",
  "nbd14",
  "nbd0",
  "nbd9",
  "nbd12",
  "nbd7",
  "nbd10",
  "nbd5"
]
```
have
```
- "nbd3"
- "nbd15"
- "nbd1"
- "nbd13"
- "nbd8"
- "nbd11"
- "nbd6"
- "nbd4"
- "nbd2"
- "nbd14"
- "nbd0"
- "nbd9"
- "nbd12"
- "nbd7"
- "nbd10"
- "nbd5"
```

erwindon:
@dawidmalina 
The Salt API is handling JSON data. Therefore SaltGUI is also using JSON data. This can be compared to the use of commandline option `--out=json` on the `salt` commandline, Indeed the `salt` command has a YAML output, unless specified otherwise.
The first layer of data is an object with the data per grain. That object is taken apart to get the presentation as it is now. There is no further interpretation on the grain details.
I fail to see a generic solution on your request here, but additional suggestions are welcome.

dawidmalina:
I know this is raw json but this exactly the same story as with state output (at the beginning we had was raw json, right?) which now is pretty formatted output. And with the the pillars or grains we usually can end up with object of type: string, list, map (last two can nest more object but also with the same type). So what I had in mind was quite simple formatter which will detect one of this types.

dawidmalina:
Maybe this library can be used: https://www.npmjs.com/package/json-to-pretty-yaml

erwindon:
I'm not worried/afraid of the actual code needed to create such presentation data.

I did not realize that the `salt` command-line is actually taking JSON parameters and producing YAML output. So there should be less reason to fear similar behaviour in SaltGUI. I'll create a prototype.

dawidmalina:
[...]

PR #127, opened 2019-01-03
UX grains information as YAML #99

dawidmalina:


erwindon:
@dawidmalina 
I will not apply this PR as is.
The change is made very specific for the grains/pillar details pages only. It fails to work on the pillar details page and is not active on any other page, which is very asymmetric.
SaltGUI is not a salt command-line simulator.
I did ask you what the purpose of this presentation change was and there was no answer. Can you please provide that answer, preferably in the description of this PR?
I am doubting about the actual purpose. Do you want to SEE data as YAML, or do you want to (RE)USE data as YAML. In the first case I will refuse it anyway since there is no obvious benefit over the existing JSON presentation as it is just a different presentation. In the second case, I would choose for an alternative implementation: Use "View as JSON" and "View as YAML" (or "Copy as JSON" and "Copy as YAML") menu items on each significant piece of data.

dawidmalina:
It's all about yaml which is designed to be readable and editable by humans where json is just a serialization format.

erwindon:
Is `json2yaml.js` your work, or is it re-use?

I'll arrange for presentation in YAML, but not as the default format. I;ll think of a way to switch representations. It will then be used in all pages. i.e. also current job output, historic job output, schedule definition (etc). then also possible to view highstate as json or yaml.

dawidmalina:
It's a fork with some modifications according jslint. Check https://github.com/dawidmalina/json2yaml.

Maybe we could just configure it via master configuration file?

erwindon:
I learnt something today...
The default output from the salt commandline is NOT `--output=yaml`, it is `--output=nested`.
It looks almost the same.

Just checking...
Is your request for NESTED of for YAML?

erwindon:
I was missing the `----------` in the yaml output... :-)

dawidmalina:
Here you go :)
zrzut ekranu 2019-01-5 o 22 16 52

erwindon:
@dawidmalina  I will ignore this pull request and go for my own implementation. That implementation is more generic and supports multiple output formats. I may still use `json2yaml.js` though.
The implementation supports:
* Output is configurable through the `master` file.
* The current (default) configuration is something like `doc,highstate,json`.
* `doc` and `highstate` are only active when the output can be identified as such.
* Other supported outputs are `yaml` and `nested`. But specifying more than one from `[json,yaml,nested]` is useless since these can always be applied.
* syntax colouring is not added.
I'll try to show a working version this week.

dawidmalina:
Ok looks like this topic will be fully covered by https://github.com/erwindon/SaltGUI/pull/133

PR #132, opened 2019-01-05
Fix package name for centos

dawidmalina:
CentOS package name is different and highstate failed:
```
local:
----------
          ID: requirements
    Function: pkg.installed
      Result: False
     Comment: The following packages failed to install/update: vim
     Started: 21:37:56.588012
    Duration: 9044.587 ms
     Changes:

Summary for local
------------
Succeeded: 0
Failed:    1
```

### DIFF
--- a/docker/srv/salt/install-packages.sls
+++ b/docker/srv/salt/install-packages.sls
@@ -1,9 +1,0 @@
-{%- if grains['os_family'] == 'Debian' %}
-  {%- set package = "vim" %}
-{%- elif grains['os_family'] == 'RedHat' %}
-  {%- set package = "vim-enhanced" %}
-{%- endif %}
-requirements:
-  pkg.installed:
-    - pkgs:
-      - {{ package }}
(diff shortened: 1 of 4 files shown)


docker/srv/salt/install-packages.sls
```diff
@@ -1,4 +1,9 @@
+{%- if grains['os_family'] == 'Debian' %}
+  {%- set package = "vim" %}
+{%- elif grains['os_family'] == 'RedHat' %}
+  {%- set package = "vim-enhanced" %}
+{%- endif %}
```
erwindon: There must be a default, which is `vim`. But in that case the statement can be even slightly simplified.

dawidmalina: `vim` is a virtual package and looks it's not working properly with CentOS (you can check by yourself). What would be your proposition according simplification? It's quite simple and follow https://docs.saltstack.com/en/latest/topics/best_practices.html (except exposing package name with context).

dawidmalina:
But you know. We are talking about fixed number of operating systems related to existing 3 minions in our demo - based on docker-compose. We don't need to cover all types so we don't need support here more then what we have to. And again we are more talking then is needed ( this is not nasa Project :)  ) - it's just saltstack state example.

erwindon:
The previous code was generic and worked for all operating systems, even the ones that were not actually included. Your change properly supports a new OS, but blocks the addition of all other OSs. And with roughly the same amount of effort it can generic with all the known exceptions still supported.

erwindon:
tested ok

PR #135, opened 2019-01-09
change ☰ to x on hover

dawidmalina:


erwindon:
I have multiple problems with this:
1) why is this useful? user can already see the menu is dropped-down. The `x` suggests an action, but no additional action becomes available. do you have a reference to a similar gui?;
2) The (css) implementation currently only works on the main screen. there should be consistency between the pages;
3) The CSS is not correctly formatted according to `stylelint`, see the PR report.

dawidmalina:
Well. I just saw few examples of such behaviour and that's it :) and it was easy to introduce. To be honest in scale from 1 to 10 I would like to have this with score 4. So if this is not desired change I can live without it ;)

And sure for the next pull request I would check and fix any issues with lint. Let me close this pull request.

PR #136, opened 2019-01-09
minimal ux improvment for: table, font, color, msg

dawidmalina:
This is the minimal set of modifications which I would really like to have. If you compare with pull request related with x for drop down menu here I would give 9:)

I was planning also change sorting icon but I have issue with additional change (some encoding issue):
```
diff --git a/saltgui/static/sorttable/sorttable.js b/saltgui/static/sorttable/sorttable.js
index 38b0fc6..e36eccb 100644
--- a/saltgui/static/sorttable/sorttable.js
+++ b/saltgui/static/sorttable/sorttable.js
@@ -100,7 +100,7 @@ sorttable = {
             this.removeChild(document.getElementById('sorttable_sortfwdind'));
             sortrevind = document.createElement('span');
             sortrevind.id = "sorttable_sortrevind";
-            sortrevind.innerHTML = stIsIE ? '&nbsp<font face="webdings">5</font>' : '&nbsp;&#x25B4;';
+            sortrevind.innerHTML = stIsIE ? '&nbsp<font face="webdings">5</font>' : '&nbsp;\u2191';
             this.appendChild(sortrevind);
             return;
           }
@@ -113,7 +113,7 @@ sorttable = {
             this.removeChild(document.getElementById('sorttable_sortrevind'));
             sortfwdind = document.createElement('span');
             sortfwdind.id = "sorttable_sortfwdind";
-            sortfwdind.innerHTML = stIsIE ? '&nbsp<font face="webdings">6</font>' : '&nbsp;&#x25BE;';
+            sortfwdind.innerHTML = stIsIE ? '&nbsp<font face="webdings">6</font>' : '&nbsp;\u2193';
             this.appendChild(sortfwdind);
             return;
           }
@@ -134,7 +134,7 @@ sorttable = {
           this.className += ' sorttable_sorted';
           sortfwdind = document.createElement('span');
           sortfwdind.id = "sorttable_sortfwdind";
-          sortfwdind.innerHTML = stIsIE ? '&nbsp<font face="webdings">6</font>' : '&nbsp;&#x25BE;';
+          sortfwdind.innerHTML = stIsIE ? '&nbsp<font face="webdings">6</font>' : '&nbsp;\u2193';
           this.appendChild(sortfwdind);

                // build an array to sort. This is a Schwartzian transform thing,
@@ -169,7 +169,7 @@ sorttable = {
     for (var i=0; i<table.tBodies[0].rows.length; i++) {
       text = sorttable.getInnerText(table.tBodies[0].rows[i].cells[column]);
       if (text != '') {
-        if (text.match(/^-?[<A3>$<A4>]?[\d,.]+%?$/)) {
+        if (text.match(/^-?[�$�]?[\d,.]+%?$/)) {
           return sorttable.sort_numeric;
         }
         // check for a date: dd/mm/yyyy or dd/mm/yy
```

Do you know what is going on?

### DIFF
--- a/docker/srv/salt/top.sls
+++ b/docker/srv/salt/top.sls
@@ -1,5 +1,3 @@
 base:
   '*':
     - vim
-    - python
-    - users
--- a/docker/srv/salt/top.sls
+++ b/docker/srv/salt/top.sls
@@ -1,3 +1,5 @@
 base:
   '*':
     - vim
+    - python
+    - users
--- a/saltgui/static/stylesheets/dropdown.css
+++ b/saltgui/static/stylesheets/dropdown.css
@@ -16,41 +16,36 @@ header .menu-dropdown-content {
 /* Dropdown Content (Hidden by Default) */
 .run-command-button .menu-dropdown-content {
   display: none;
-  box-shadow: 0 8px 16px 0 rgba(0, 0, 0, 0.5);
   position: absolute;
   max-height: 210px;
   overflow-y: auto;
   overflow-x: visible;
-  background-color: #e0e0e0;
-  border: 10px solid #e0e0e0;
+  background: #fff;
+  box-shadow: 0 3px 10px -2px rgba(0, 0, 0, 0.1);
+  border: 1px solid rgba(0, 0, 0, 0.1);
   z-index: 5;
   cursor: pointer;
 }
 
 /* Links inside the menu-dropdown */
 .run-command-button .menu-dropdown-content div {
-  display: inline-block;
+  display: block;
   float: left;
   clear: both;
-  margin: 3px;
-  padding: 5px 20px 5px 20px;
-
-  /* correct width for margin-left/right,padding-left/right */
-  width: calc(100% - 20px - 20px - 3px - 3px);
+  width: 100%;
+  padding: 15px 30px 15px 30px;
+  -webkit-transition: all 0.2s ease-in-out;
+  -moz-transition: all 0.2s ease-in-out;
+  transition: all 0.2s ease-in-out;
 }
 
 /* Change color of menu-dropdown links on hover */
 .run-command-button .menu-dropdown-content div:hover {
-  background-color: #efefef;
+  background: rgba(0, 0, 0, 0.15);
   cursor: pointer;
 }
 
 /* Show the menu-dropdown menu on hover */
 .run-command-button:hover .menu-dropdown-content {
   display: block;
-}
-
-/* Change the background color of the menu-dropdown button when the menu-dropdown content is shown */
-.run-command-button:hover .menu-dropdown {
-  background-color: #efefef;
 }
--- a/saltgui/static/stylesheets/dropdown.css
+++ b/saltgui/static/stylesheets/dropdown.css
@@ -17,6 +17,7 @@ header .menu-dropdown-content {
 .run-command-button .menu-dropdown-content {
   display: none;
   position: absolute;
+  max-height: 285px;
   overflow-y: auto;
   overflow-x: visible;
   background: #fff;
--- a/saltgui/static/stylesheets/dropdown.css
+++ b/saltgui/static/stylesheets/dropdown.css
@@ -17,7 +17,7 @@ header .menu-dropdown-content {
 .run-command-button .menu-dropdown-content {
   display: none;
   position: absolute;
-  max-height: 285px;
+  max-height: 300px;
   overflow-y: auto;
   overflow-x: visible;
   background: #fff;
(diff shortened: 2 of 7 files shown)


dawidmalina:
On my side it's looks like this:
zrzut ekranu 2019-01-9 o 23 48 24

erwindon:
In the 4th div section, a string of 3 characters exist.
These are the 3 monetary symbols: pound(A3), dollar and currency-sign(A4). These byte-codes exist in "extended-ascii". When you open the file in an editor that assumes e.g. UTF-8, the characters change their bytes to another character set.
That is why we use \\u0000 and &#0000 notation nowadays.

dawidmalina:
Thx, change is now finished. Please review :)

erwindon:
looks nice, some points...
* I'm missing the highlight on the menu character when the drop-down menu shows
* The menu for small screens (<800px) has now become too small and it now uses a scroll bar.
* Is there a (subtle) background shade that can be used inside the drop-down menu? that shows the menu more separate from the background. Previously that was done with a shadow.

dawidmalina:
I will looks on your point tomorrow. The only thing which is missing for me is: "table row is highlighted on mouseover". Do you think you would accept this?

dawidmalina:
> I'm missing the highlight on the menu character when the drop-down menu shows

This is because I still believed that row highlight will be accepted;) So entire row will be highlighted.

> The menu for small screens (<800px) has now become too small and it now uses a scroll bar.

How do you testing this? I've checked only with firebug and iPad screen size (768 x 1024) and I didn't found any issues

> Is there a (subtle) background shade that can be used inside the drop-down menu? that shows the menu more separate from the background. Previously that was done with a shadow.

But there is a shadow for entire dropdown content
```
  box-shadow: 0 3px 10px -2px rgba(0, 0, 0, 0.1);
  border: 1px solid rgba(0, 0, 0, 0.1);
```

erwindon:
> How do you testing this?
Un-maximize the browser window. Then drag the sides to reduce the width.
afbeelding


The shadow is so faint with alpha=0.1. My minimum is 0.3.

dawidmalina:
Thx for the hint. Now should looks much better.
zrzut ekranu 2019-01-12 o 06 36 51

saltgui/static/stylesheets/dropdown.css
```diff
@@ -17,7 +17,6 @@ header .menu-dropdown-content {
 .run-command-button .menu-dropdown-content {
   display: none;
   position: absolute;
-  max-height: 210px;
```
erwindon: When you remove this, generated menu's can become too large. Currently the "templates" menu is generated. It is located in the command-box, just right of the words "Manual Run". Any menu is only shown when there are items in the menu. In this case that needs configuration in the `master` file. With the `max-height`, a menu with scrollbar is produced when there are too many menuitems.

dawidmalina: Ok. Let me then bring back `max-height` with proper size to address existing top menu. And maybe "template" could be moved to dedicated page where all templates could be easily displayed and executed. I can work on this just after dedicate job page.

dawidmalina: And according templates let me add (different pull request) sample templates to configuration in docker - this could help during testing and can be useful for new users to explore all SalGUI features.

docker/srv/salt/top.sls
```diff
@@ -1,3 +1,5 @@
 base:
   '*':
     - vim
+    - python
+    - users
```
erwindon: why was this introduced?

dawidmalina: Wow. Sorry about this. It's my local change to test highstate output. I will fix this. Good catch :)

saltgui/static/stylesheets/dropdown.css
```diff
@@ -17,6 +17,7 @@ header .menu-dropdown-content {
 .run-command-button .menu-dropdown-content {
```
erwindon: I found an even better solution. The main menu (on narrow screens) should always have the size it wants, independent of any other restrictions.
Can you please include the following below this one:
```css
#mainmenumini .run-command-button .menu-dropdown-content {
  max-height: none;
}
```
Note: it should become (almost) the last entry in the file to keep `stylelint` happy.

dawidmalina: Added. Thx for hint.

erwindon:
Additionally created a new release number.
Please verify the README.MD to see whether I missed anything or made any errors there

dawidmalina:
Nice :+1: there might be one small change. I think it should be `Fixed issues with ip-number when multiple available` instead of `Fixed issues when ip-number when multiple available`. Rest is good!

PR #137, opened 2019-01-12
Add simple templates configuration

dawidmalina:
This could help during testing and can be useful for new users to explore all SaltGUI features


erwindon:
thx

Issue #116, opened 2018-12-31
UX recent jobs improvements


dawidmalina:
It would be nice improvement if we could really see status of recent jobs. What we could do is limit the number of jobs that are treated as recent and fetch for this jobs detailed information by executing sub request `jobs.list_jobs_filter 10` and then `jobs.list_job xxxxxxxx`

dawidmalina:
I totally agree with you - as I didn't thought about multiple operators in the same time.

erwindon:
See issue #180 and PR #205

Issue #63, opened 2018-10-03
Commandbox should get its own route and page


maerteijn:
Now it's somehow a strange thing inside of SaltGUI. It would make much more sense to create a own route / page for it as the functionality for it is growing.

erwindon:
This would eliminate the problem of updating the underlying screen whenever a command is run in the commandbox. On the other hand, the commandbox is used for just these cases. e.g. accepting a key on the keys page, etc. you don't want to manually switch back each time to see what to do next (accepting the next key, applying the next state, etc). So, some kind of "go back" feature should remain. Maybe just the 'X' that is already on the current screen.
Interesting to know what the fate of the "green circle" will be...

maerteijn:
I don’t understand what you mean with “you don’t want to manually switch back to see what yo do next”

erwindon:
The new command-page is slightly different from other pages. This is because you want to go back to the location where you initiated the navigation to the command-page. With a popup-panel (like now) that function is done with the 'X' in the top-right corner. When it is a page, something similar must remain. It then does not 'close' the panel, but navigates to the previous page.

maerteijn:
Ah yes, that is something you want  indeed. I'll try some things first to see if  this idea makes sense.

maerteijn:
I'll leave it up to you to close this

erwindon:
no better suggestions have popped up

PR #138, opened 2019-01-13
New dedicated jobs page

dawidmalina:
Let me introduce new dedicated jobs page where we will show (currently 20 recent jobs - but this later could be extend for all jobs with pagination) recent jobs.

Then each job will have a drop down menu to direct to job details (result).

Regular view:
zrzut ekranu 2019-01-13 o 20 45 50

Small screen view:
zrzut ekranu 2019-01-13 o 20 46 04

Later we can even introduce job status (as you already have some prototype - https://github.com/erwindon/SaltGUI/issues/116) and maybe we could do something with nanoseconds (as I think they are not needed in list view). I was also thing about new drop down entry to rerun particular job. What do you think about this topics?

Btw you still didn't answer me whether you will accept table row color change on hover, as this is what I still missing ;)

### DIFF
--- a/saltgui/index.html
+++ b/saltgui/index.html
@@ -80,6 +80,10 @@
               </tbody>
             </table>
           </div>
+          <div class='panel job-list'>
+            <h1>Recent Jobs</h1>
+            <table class='jobs'></table>
+          </div>
         </div>
       </div>
 
@@ -95,6 +99,10 @@
               </tbody>
             </table>
           </div>
+          <div class='panel job-list'>
+            <h1>Recent Jobs</h1>
+            <table class='jobs'></table>
+          </div>
         </div>
       </div>
 
@@ -109,6 +117,10 @@
               <tbody>
               </tbody>
             </table>
+          </div>
+          <div class='panel job-list'>
+            <h1>Recent Jobs</h1>
+            <table class='jobs'></table>
           </div>
         </div>
       </div>
@@ -126,6 +138,10 @@
               </tbody>
             </table>
           </div>
+          <div class='panel job-list'>
+            <h1>Recent Jobs</h1>
+            <table class='jobs'></table>
+          </div>
         </div>
       </div>
 
@@ -140,6 +156,10 @@
               <tbody>
               </tbody>
             </table>
+          </div>
+          <div class='panel job-list'>
+            <h1>Recent Jobs</h1>
+            <table class='jobs'></table>
           </div>
         </div>
       </div>
@@ -157,6 +177,10 @@
               </tbody>
             </table>
           </div>
+          <div class='panel job-list'>
+            <h1>Recent Jobs</h1>
+            <table class='jobs'></table>
+          </div>
         </div>
       </div>
 
@@ -171,6 +195,10 @@
               <tbody>
               </tbody>
             </table>
+          </div>
+          <div class='panel job-list'>
+            <h1>Recent Jobs</h1>
+            <table class='jobs'></table>
           </div>
         </div>
       </div>
@@ -188,6 +216,10 @@
               </tbody>
             </table>
           </div>
+          <div class='panel job-list'>
+            <h1>Recent Jobs</h1>
+            <table class='jobs'></table>
+          </div>
         </div>
       </div>
 
@@ -202,8 +234,8 @@
 
       <div class='route' id='page_jobs'>
           <div class='dashboard'>
-            <div id='jobs_page' class='panel jobs-list'>
-              <h1 id="jobs_title">Recent Jobs</h1>
+            <div id='jobs_page' class='panel jobs-list' style="margin-right: 0;">
+              <h1 id="jobs_title">Last 20 Jobs</h1>
               <table id="jobs" class='jobs sortable'>
                 <thead>
                 <tr><th>JID</th><th>Target</th><th>Function</th><th>Start Time</th><th class="sorttable_nosort"></th></tr>
--- a/saltgui/index.html
+++ b/saltgui/index.html
@@ -234,8 +234,8 @@
 
       <div class='route' id='page_jobs'>
           <div class='dashboard'>
-            <div id='jobs_page' class='panel jobs-list' style="margin-right: 0;">
-              <h1 id="jobs_title">Last 20 Jobs</h1>
+            <div id='jobs_page' class='panel jobs-list'>
+              <h1 id="jobs_title">Recent Jobs</h1>
               <table id="jobs" class='jobs sortable'>
                 <thead>
                 <tr><th>JID</th><th>Target</th><th>Function</th><th>Start Time</th><th class="sorttable_nosort"></th></tr>
@@ -252,7 +252,7 @@
     <div class='popup' id='popup_runcommand'>
       <div class='run-command'>
         <div class='nearlyvisiblebutton' id='button_close_cmd'>&#x2715;</div>
-        <div><h1>Manual Run</h1><p style="display: inline;" id="templatemenuhere"></p></div>
+        <div><h1>Manual Run</h1><p id="templatemenuhere"></p></div>
         <div id="targetbox"><input list="targetlist" id='target' type='text' placeholder="Target"/><datalist id="targetlist"/></div>
         <div id="cmdbox"><input id='command' type='text' placeholder="Command"/></div>
         <div id="runblock"><input id="run_command" type='submit' value="Run command"/></div>
--- a/saltgui/index.html
+++ b/saltgui/index.html
@@ -82,7 +82,7 @@
           </div>
           <div class='panel job-list'>
             <h1>Recent Jobs</h1>
-            <table class='jobs'></table>
+            <table class='jobs left-panel'></table>
           </div>
         </div>
       </div>
@@ -101,7 +101,7 @@
           </div>
           <div class='panel job-list'>
             <h1>Recent Jobs</h1>
-            <table class='jobs'></table>
+            <table class='jobs left-panel'></table>
           </div>
         </div>
       </div>
@@ -120,7 +120,7 @@
           </div>
           <div class='panel job-list'>
             <h1>Recent Jobs</h1>
-            <table class='jobs'></table>
+            <table class='jobs left-panel'></table>
           </div>
         </div>
       </div>
@@ -140,7 +140,7 @@
           </div>
           <div class='panel job-list'>
             <h1>Recent Jobs</h1>
-            <table class='jobs'></table>
+            <table class='jobs left-panel'></table>
           </div>
         </div>
       </div>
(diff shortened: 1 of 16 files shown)


erwindon:
> Btw you still didn't answer me whether you will accept table row color change on hover, as this is what I still missing ;)
Not ignoring you, the topic was just lost in my mind...
Table row colour change (on hover) is only acceptable when there is no other indication. This happens with the current side-panel for jobs, which does not have a drop-down menu. And as I see it that will also soon change. So, not really an acceptable item.

I already had the visual removal of nanoseconds in mind. Please do not combine it with this PR, but raise as a separate issue.

For jobs I did not have the visual changes in mind yet. But giving it a normal drop-down menu was alreay planned. First of all to give it the same feel as the left panel, with the menu item "Show details...". Next was the addition of (indeed) a "Rerun job..." facility. That one exists in multiple forms: "re-run-all", "re-run-unreachable", "re-run-failed" and "re-run-unsuccesfull" (the combination of the previous 2). But the problem still remains about having the details available. So these actions might have to be part of the details view itself.

Please take care of placing large PRs without discussing them first. I might (in general) not honour the request, or prefer to treat it as an idea for which I want my own implementation instead.

A facility must be present to choose which "target" values are too long for display. The target might be a huge list of minion names (e.g. as generated by some retry facility, see above).

Where did the output of `jobs.active` go to? It used to show the number of targeted minions and the number of minions that returned... That should be something like a "Progress" column.

Paging should indeed not be included in this PR.

Jobs as a separate page will have a fundamental problem. The jobs list on the right-hand side of the windows is quite useful to see whether commands have finished. I don't mind having a new page with more control on the jobs, But actually, I want to keep the small job list too.

dawidmalina:
1. done - https://github.com/erwindon/SaltGUI/issues/140
2. done - "Show details..." introduced and rerun should be added in separated pr - https://github.com/erwindon/SaltGUI/issues/141
3. ack - will do this next time - that's why I've asked you about slack :)
4. do you have sth in mind how this can be controlled?
5. should be also on the list - I will double check this and introduce status
6. sure https://github.com/erwindon/SaltGUI/issues/142
7. so my proposition looks like this. having this side window does not bring imho any value on all pages except of run command page where indeed having this could be valuable. this could be part of changes related with this: https://github.com/erwindon/SaltGUI/issues/63

erwindon:
4. just a simple CSS width restriction should do the trick

7. I think the jobs side-panel is useful on almost al pages as it shows you the history of commands. No need to switch back-and-forth between the information page and the jobs page. Note that this is a typical difference between task-oriented interfaces and command-orientated interfaces. I like the former.

erwindon:
Yes, that is the bug that is already known.
The details are correct while the job is still running. So running `test.rand_sleep 60` (async) should return full details with `jobs.list_jobs`

saltgui/index.html
```diff
@@ -200,8 +232,8 @@ <h2 class='time'></h2>
 
       <div class='route' id='page_jobs'>
           <div class='dashboard'>
-            <div id='jobs_page' class='panel jobs-list'>
-              <h1 id="jobs_title">Recent Jobs</h1>
+            <div id='jobs_page' class='panel jobs-list' style="margin-right: 0;">
+              <h1 id="jobs_title">Last 20 Jobs</h1>
```
erwindon: I would still call it `Recent Jobs`. Because there might not be 20 visible, and we are making a selection on the jobs that we show.

saltgui/static/stylesheets/page.css
```diff
@@ -101,6 +100,14 @@ table tr td:last-of-type {
   padding-right: 0;
 }
 
+table tr td.starttime,
+table tr td.target {
+  white-space: pre-wrap;
+  max-width: 20%;
+  width: 20%;
```
erwindon: `width` should not be neccesary?

saltgui/static/stylesheets/page.css
```diff
@@ -101,6 +100,14 @@ table tr td:last-of-type {
...
+  display: table-cell;
```
erwindon: Isn't it already a `table-cell`?

saltgui/static/scripts/routes/page.js
```diff
@@ -216,6 +217,127 @@ class PageRoute extends Route {
...
+      if(job.Function === "schedule.list") continue;
+      if(job.Function === "schedule.modify") continue;
+      if(job.Function === "schedule.run_job") continue;
+      if(job.Function === "sys.doc") continue;
+      if(job.Function === "wheel.config.values") continue;
+      if(job.Function === "wheel.key.accept") continue;
+      if(job.Function === "wheel.key.delete") continue;
+      if(job.Function === "wheel.key.list_all") continue;
+      if(job.Function === "wheel.key.reject") continue;
+
```
erwindon: This block is now duplicated. The part with the logic on which jobs to show (or not) should be separated.

saltgui/index.html
```diff
@@ -200,8 +232,8 @@ <h2 class='time'></h2>
 
       <div class='route' id='page_jobs'>
           <div class='dashboard'>
-            <div id='jobs_page' class='panel jobs-list'>
-              <h1 id="jobs_title">Recent Jobs</h1>
+            <div id='jobs_page' class='panel jobs-list' style="margin-right: 0;">
```
erwindon: style item should move to css file

saltgui/static/stylesheets/page.css
```diff
@@ -101,6 +100,14 @@ table tr td:last-of-type {
   padding-right: 0;
 }
 
+table tr td.starttime,
```
erwindon: what is the reason to apply this to starttime?

saltgui/index.html
dawidmalina: Will revert this.

saltgui/static/stylesheets/page.css
dawidmalina: Without `width` container is always shrinked to the minimal width and wrapping is applied. That's why I've added this.

saltgui/static/stylesheets/page.css
dawidmalina: Yes, it's not required. I will remove it.

saltgui/static/scripts/routes/page.js
dawidmalina: Sure. Will do this.

saltgui/index.html
dawidmalina: Sure. Will do this. I will fixed also `<p style="display: inline;" id="templatemenuhere">`

saltgui/static/stylesheets/page.css
dawidmalina: I was found this useful to shrink also date depending on windows size:
- before
```
2019, Jan 17 23:29:07.065324
```
- after
```
2019, Jan 17 
23:29:07.065324
```

But I can remove it if you don't like it.

dawidmalina:
@erwindon can you please review?

saltgui/static/stylesheets/page.css
erwindon: When you use `white-space: nowrap`, then `width` and `max-width` can be removed.

saltgui/static/scripts/routes/job.js
```diff
@@ -20,7 +20,7 @@ class JobRoute extends Route {
     job.getPageElement().querySelector(".output").innerText = "";
 
     document.querySelector("#button_close_job").addEventListener("click", _ => {
-      this.router.goTo("/");
+      this.router.goTo("/jobs");
```
erwindon: Can this be re-organized to go back to the page where the user came from?
One can end up on the jobs page directly from the home page, so you want to return to the home page in that case.

erwindon:
Can you REBASE this branch?
(reminder: do not merge `master` into the branch)
Feel free to collapse trivial changes such as `Add missing semicolon` or `Remove console.log`.

saltgui/static/scripts/routes/job.js
dawidmalina: Thx for the hint. I've ended up with this: `window.location.reload(history.back());`

saltgui/static/stylesheets/page.css
dawidmalina: `width` and `max-width` removed.

dawidmalina:
Branch rebased and `Add missing semicolon` and `Remove console.log` commit squashed.

dawidmalina:
Having this `window.location.reload(window.history.back());` work should allow us to go further with this: https://github.com/erwindon/SaltGUI/issues/63 where one of the problem mentioned by you was how to go back to previous page from run command dedicated page.

dawidmalina:
Hope those were the last comments and we can merge this :)

saltgui/static/scripts/routes/job.js
erwindon: Unfortunately, that does not work...
(also: `history.back` does not have a return value and `location.reload` does not take a parameter. it should have been 2 separate commands, but that does not change the situation)
The current navigation framework uses a mix of `location.assign` and `location.replace`, combined with explicit `history.pushState` manipulations. Let's not mess with that now. The current version already has the same problem when showing job details from most pages.
Let's return to your original code of using `/jobs'. Sorry for the detour.

dawidmalina: Ok. I will revert this part - strange as this was really working in my case (firefox browser).

erwindon: I had the same conclusion about the combined code and also used the newly suggested code. Now retried it again, but no success.
It works on all pages, except for job details reached from the jobs page itself.

erwindon: SaltGUI is mostly using `location.replace`, thus destroying the history. See also https://www.w3schools.com/jsref/met_loc_replace.asp. Let's not rely on `windows.history`.

dawidmalina: I found out why it was not working on jobs page. On most pages we are going directly from particular page to job details (so `history.go(-1)` is enough) however on jobs page we are going through event callback (so `history.go(-1)` is not enough and it will work only with `history.go(-2)` ).

So I was able to workaround this by this code:
```
    document.querySelector("#button_close_job").addEventListener("click", _ => {
      window.history.back();
      // when we were in jobs page we need to bo back twice
      if (window.location.pathname == '/job') {
        window.history.back();
      } else {
        // on rest of the pages we need to reload page
        window.location.reload();
      }
    });
```

dawidmalina: Anyway code reverted to `this.router.goTo("/jobs");`

erwindon:
no need to push.
it already has the highest priority.
I expect this to happen today.

erwindon:
code cleanup (moving jobs related code to `jobs.js`) added as issue #153
missing functionality of jobs page added as issue #154

PR #151, opened 2019-01-27
removal of nanoseconds in job start time

erwindon:
See #151 
Removal of nanoseconds in job start time does not bring any value and I think that can be removed.


erwindon:
@dawidmalina can you please review/test this change?

dawidmalina:
Yes, it's working. At the very beginning I was looking for just an option to remove this nano part. If you feel that ability to configure precision is better (not just remove or keep) - you have my approve 👍

PR #157, opened 2019-01-29
dedicated job template page

### LINKED ISSUE #145: dedicated job template page
It would be nice to have dedicate job template page where all configured templates will be displayed.
In addition page will allows us to apply particular template (forward to run command with needed inputs filled up automatically).

This is proposition (mockup) which I can work on:
zrzut ekranu 2019-01-20 o 20 43 41

When user will not have any templates message `No job template found` should be displayed instead.

dawidmalina:
closes #145

### DIFF
--- a/saltgui/static/scripts/routes/page.js
+++ b/saltgui/static/scripts/routes/page.js
@@ -29,7 +29,7 @@ class PageRoute extends Route {
         window.location.replace("/jobs");
       });
       // hide template menu item if no templates defined
-      let templatesText = window.localStorage.getItem("templates");
+      const templatesText = window.localStorage.getItem("templates");
       if (templatesText && templatesText !== "undefined") {
         menu.addMenuItem("templates", function(evt) {
           window.location.replace("/templates");
--- a/saltgui/static/scripts/routes/page.js
+++ b/saltgui/static/scripts/routes/page.js
@@ -303,7 +303,6 @@ class PageRoute extends Route {
     tr.appendChild(td);
 
     td.id = "job" + job.id;
-
     const targetText = window.makeTargetText(job["Target-type"], job.Target);
     td.appendChild(Route._createDiv("target", targetText));
 
@@ -320,7 +319,6 @@ class PageRoute extends Route {
 
   _addDetailedJob(container, job) {
     const tr = document.createElement("tr");
-
     const jidText = job.id;
     tr.appendChild(Route._createTd("jid", jidText));
 
--- a/saltgui/static/scripts/routes/page.js
+++ b/saltgui/static/scripts/routes/page.js
@@ -387,7 +387,11 @@ class PageRoute extends Route {
     document.execCommand("copy");
   }
 
-  _runCommand(evt, targetString, commandString, targettype) {
+  _runCommand(evt, targetString, commandString) {
+    this._runFullCommand(evt, "", targetString, commandString);
+  }
+
+  _runFullCommand(evt, targettype, targetString, commandString) {
     this.router.commandbox._showManualRun(evt);
     const target = document.querySelector("#target");
     const command = document.querySelector("#command");
--- a/saltgui/static/scripts/routes/page.js
+++ b/saltgui/static/scripts/routes/page.js
@@ -28,6 +28,13 @@ class PageRoute extends Route {
       menu.addMenuItem("jobs", function(evt) {
         window.location.replace("/jobs");
       });
+      // hide template menu item if no templates defined
+      let templatesText = window.localStorage.getItem("templates");
+      if (templatesText && templatesText !== "undefined") {
+        menu.addMenuItem("templates", function(evt) {
+          window.location.replace("/templates");
+        });
+      }
       menu.addMenuItem("logout", function(evt) {
         const api = new API();
         api.logout().then(window.location.replace("/"));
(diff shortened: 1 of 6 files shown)


erwindon:
please add text `closes #145` in the description of this PR so that GitHub knows which issues to close after PR has been merged

erwindon:
I think the column "Target Type" can be removed from the screen.
When it has a value, it can simply be placed before the target value.

"should we update it or just use from cache" --> update is ok

"fill out the number of columns to that of the header" --> that code should not be needed here. this page has a fixed number of columns and the code is providing values for all columns

saltgui/static/scripts/routes/page.js
```diff
@@ -380,10 +387,26 @@ class PageRoute extends Route {
     document.execCommand("copy");
   }
 
-  _runCommand(evt, targetString, commandString) {
+  _runCommand(evt, targetString, commandString, targettype) {
```
erwindon: please always use the sequence "targettype, targetvalue", in that order, and always next to each other

saltgui/static/scripts/routes/templates.js
```diff
@@ -0,0 +1,85 @@
...
+    if (!target) {
+      tr.appendChild(Route._createTd("target none", "(none)"));
+    } else {
+      tr.appendChild(Route._createTd("target", target));
+    }
+
+    // calculate targettype
+    let targettype = template["targettype"];
+    if (!targettype) targettype = "";
+    tr.appendChild(Route._createTd("targettype", targettype));
```
erwindon: I think the column "Target Type" can be removed from the screen.
When it has a value, it can simply be placed before the target value.

saltgui/static/scripts/routes/templates.js
```diff
@@ -0,0 +1,85 @@
...
+      templates.router.api.getConfigValues().then(templates._updateTemplates);
+    });
+  }
+
+  _updateTemplates(data) {
+    const container = this.getPageElement().querySelector(".templates");
+    
+    // should we update it or just use from cache (see commandbox) ?
+    const templates = data.return[0].data.return.saltgui_templates;
+    window.localStorage.setItem("templates", JSON.stringify(templates));
```
erwindon: updating the cache here is ok

saltgui/static/scripts/routes/templates.js
```diff
@@ -0,0 +1,85 @@
...
+    const command = template["command"];
+    if (!command) {
+      tr.appendChild(Route._createTd("command none", "(none)"));
+    } else {
+      tr.appendChild(Route._createTd("command", command));
+    }
+
+    const menu = new DropDownMenu(tr);
+    menu.addMenuItem("Apply&nbsp;template...", function(evt) {
+      this._applyTemplate(evt, template["target"], template["targettype"], template["command"]);
```
erwindon: these values are already retrieved in simple variables, why not use these instead?

saltgui/static/scripts/routes/templates.js
```diff
@@ -0,0 +1,85 @@
...
+    }.bind(this));
+
+    // fill out the number of columns to that of the header
+    while(tr.cells.length < container.tHead.rows[0].cells.length) {
+      tr.appendChild(Route._createTd("", ""));
+    }
```
erwindon: that code should not be needed here. this page has a fixed number of columns and the code is providing values for all columns

saltgui/static/stylesheets/page.css
```diff
@@ -106,6 +109,14 @@ table tr td.target {
   white-space: nowrap;
 }
 
+.none {
+  opacity: 0.4;
+}
+
+.disabled {
```
erwindon: please rename to `menuitem_hidden`

saltgui/static/stylesheets/page.css
```diff
@@ -106,6 +109,14 @@ table tr td.target {
   white-space: nowrap;
 }
 
+.none {
```
erwindon: please rename to `.value_none`

saltgui/static/scripts/routes/templates.js
dawidmalina: This is what I would like to debate with you:). I think that it's useful to have this information id dedicated column as the aim of this is to show (exact) what we have under master configuration and here we have dedicated field `targettype`.

saltgui/static/stylesheets/page.css
dawidmalina: 👍

saltgui/static/stylesheets/page.css
dawidmalina: 👍

saltgui/static/scripts/routes/templates.js
dawidmalina: 👍

saltgui/static/scripts/routes/templates.js
dawidmalina: 👍

saltgui/static/scripts/routes/page.js
dawidmalina: 👍

saltgui/static/scripts/routes/templates.js
dawidmalina: Ok. I've merged this as you mentioned as it doesn't look so bad :)

erwindon:
I will merge it like it is now. But I will also update some details of the handling of targettype+target. Too much work to explain the details that I have in mind...

Issue #154, opened 2019-01-27
new jobs page does not show statistics on running jobs


erwindon:
**Is your feature request related to a problem? Please describe.**
The jobs summary on many pages shows for which jobs the minions are still working. This functionality is not present on the jobs page

**Describe the solution you'd like**
Add the jobs summary as a new column on the jobs page.

**Describe alternatives you've considered**
None

**Additional context**
N/A

dawidmalina:
If you will give me more details what exactly you would like to have I can work on this.

erwindon:
The jobs panel shows 7 jobs with 3 values: target, function and starttime.
For jobs that are still running it shows how many minions are already done and how many are still running the command.
The info is retrieved last, and it may take a few seconds before it is shown.
Here is a sample:
![afbeelding](https://user-images.githubusercontent.com/3663742/51804982-f510b700-2267-11e9-91f6-b9a08a3f0549.png)

The suggested solution for the Jobs page is to add a column right of the dropdownmenu button that shows this. Column title is "Status". Do not combine with the target info as done in the jobs panel.

dawidmalina:
Currently only minions and keys page run this `minions.router.api.getJobsActive().then(minions._runningJobs);` onShow. I've added this to jobs page but I didn't add this to page content. I think that this should be added to all pages to be consistent. @erwindon what do you think?

erwindon:
yes, the side-panel for jobs should behave identical wherever it is located.

dawidmalina:
It will look like this:
![zrzut ekranu 2019-01-31 o 01 17 39](https://user-images.githubusercontent.com/5910851/52021707-64033f80-24f6-11e9-9906-923cf2ed8bb2.png)

erwindon:
will be closed automatically when the PR is done

PR #162, opened 2019-01-31
Job details

dawidmalina:
closes https://github.com/erwindon/SaltGUI/issues/154

### DIFF
--- a/saltgui/index.html
+++ b/saltgui/index.html
@@ -239,7 +239,7 @@
             <h1 id="jobs_title">Recent Jobs</h1>
             <table id="jobs" class='jobs sortable'>
               <thead>
-              <tr><th>JID</th><th>Target</th><th>Function</th><th>Start Time</th><th>Status</th><th class="sorttable_nosort"></th></tr>
+              <tr><th>JID</th><th>Target</th><th>Function</th><th>Start Time</th><th class="sorttable_nosort"></th><th>Status</th></tr>
               </thead>
               <tbody>
               </tbody>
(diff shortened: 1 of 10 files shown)


saltgui/index.html
```diff
@@ -239,7 +239,7 @@ <h2 class='time'></h2>
             <h1 id="jobs_title">Recent Jobs</h1>
             <table id="jobs" class='jobs sortable'>
               <thead>
-              <tr><th>JID</th><th>Target</th><th>Function</th><th>Start Time</th><th class="sorttable_nosort"></th></tr>
+              <tr><th>JID</th><th>Target</th><th>Function</th><th>Start Time</th><th>Status</th><th class="sorttable_nosort"></th></tr>
```
erwindon: I think the status column should be right of the menu item because it is usually empty

saltgui/static/scripts/routes/page.js
```diff
@@ -296,6 +296,25 @@ class PageRoute extends Route {
...
+      if(job.Running.length > 0)
+        targetText = job.Running.length + " running";
+      if(job.Returned.length > 0)
+        targetText = targetText + ", " + job.Returned.length + " returned";
+
+      const targetField = document.querySelector(".jobs #job" + k + " .status");
+      // the field may not (yet) be on the screen
+      if(targetField) targetField.innerText = targetText;
+    }
+  }
```
erwindon: this function is so similar to `_runningJobs`. can it not be combined?

Issue #158, opened 2019-01-30
UX top menu improvments


dawidmalina:
**Is your feature request related to a problem? Please describe.**
Currently we have quite a lot of top menu items.

**Describe the solution you'd like**
It would be good if we could merge some items within common groups.
```
| minions    | keys | jobs      | logout |
| grains     |      | templates |
| schedulers |
| pillars    |
```
> menu for small screen should remain as is

**Describe alternatives you've considered**
In addition we could also move run command icon to regular item
```
| minions    | keys | jobs      | execute command | logout |
| grains     |      | templates |
| schedulers |
| pillars    |
```

**Additional context**
It would be nice if we could merge menu layout to looks similar to regular drop down for different details / actions.

PR #163, opened 2019-01-31
WIP ux top menu improvments

dawidmalina:
closes https://github.com/erwindon/SaltGUI/issues/158


dawidmalina:
I left run commend icon as is. I've tried remove it from top menu but then layout for small screen was broken as menu was not fully visible.

saltgui/static/stylesheets/main.css
```diff
@@ -19,14 +19,17 @@ header {
   border-top: 2px solid #4caf50;
   background-color: white;
   margin-top: 0;
-  padding: 12px 24px 12px 24px;
+  height: 51px;
```
erwindon: That fixed height should not be needed. It is a bit arbitrary value. Also it causes a gap between the menu title item and the dropdown panel. It looks much better when I disable it.

Issue #141, opened 2019-01-15
add ability to rerun the job


dawidmalina:
I would be nice if we could rerun the particular job.

erwindon:
First regular re-run should be added.
Its action is based on the output from `jobs.list_jobs`, which we already have.
The function opens the commandbox pre-filled with the old values.
`Function`+`Arguments` should be placed the the command field as valid commandline with json arguments. Preferably simple strings become un-quoted.
`Target-type` should be used to pre-select the target type when it is one of the 4 known targettypes. Otherwise the targettype becomes `automatic`.
`Target` should be used to pre-fill the target field.
`StartTime`+`User` are not used.

dawidmalina:
There is also one issue as currently returned tergettype is unknown (https://github.com/erwindon/SaltGUI/issues/103). So users with wrong salt version will be affected by this bug.

Another small challenge will be with arguments as sometimes it can be empty, string or object (even if argument was provided as string).

PR #164, opened 2019-01-31
add ability to rerun the job

dawidmalina:
closes https://github.com/erwindon/SaltGUI/issues/141

### DIFF
--- a/saltgui/static/scripts/routes/page.js
+++ b/saltgui/static/scripts/routes/page.js
@@ -325,7 +325,8 @@ class PageRoute extends Route {
     const targetText = window.makeTargetText(job["Target-type"], job.Target);
     tr.appendChild(Route._createTd("target", targetText));
 
-    const functionText = job.Function;
+    const argumentsText = this._decodeArgumentsText(job.Arguments[0]);
+    const functionText = job.Function + " " + argumentsText;
     tr.appendChild(Route._createTd("function", functionText));
 
     const startTimeText = job.StartTime;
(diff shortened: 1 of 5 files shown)


dawidmalina:
It is possible to re-run job from jobs page and job detail page.

saltgui/static/scripts/routes/route.js
```diff
@@ -39,4 +39,62 @@ class Route {
...
+    }
+
+    target.value = targetString;
+    command.value = commandString;
+    // the menu may become (in)visible due to content of command field
+    this.router.commandbox.cmdmenu.verifyAll();
+  }
+
+  _decodeArgumentsText(rawArguments) {
+    console.log(rawArguments);
```
erwindon: please remove this one

saltgui/static/scripts/routes/route.js
```diff
@@ -39,4 +39,62 @@ class Route {
...
+    const rawObject = rawArguments;
+    let argumentsText = "";
+
+    switch (typeof rawObject) {
+    case "undefined":
+      argumentsText = "";
+      break;
+    case "object": {
```
erwindon: does anything bad happen without the { ... } ?

saltgui/static/scripts/routes/page.js
```diff
@@ -335,6 +336,9 @@ class PageRoute extends Route {
     menu.addMenuItem("Show&nbsp;details", function(evt) {
       window.location.assign("/job?id=" + encodeURIComponent(job.id));
     }.bind(this));
+    menu.addMenuItem("Re-run&nbsp;job...", function(evt) {
+      this._runFullCommand(evt, job["Target-type"], job.Target, functionText);
```
erwindon: can you intercept the value "unknown-target".
In that case, Target-type should be default and Target should be empty.
Note that the command parameters are also lost in that case

saltgui/static/scripts/routes/route.js
dawidmalina: Yes. I don't remember exact message but eslint was failing without it.

saltgui/static/scripts/routes/page.js
dawidmalina: Ok I will try to handle "unknown-target" case and use default values it such situation.

dawidmalina:
For the remaining types of re-run (which you described) I propose to do it with dedicated pull request.

PR #165, opened 2019-02-01
add more retry options

erwindon:
Adding more menu items on the job details menu, with:
* redo all jobs that did not have a response, not visible when n/a
* redo all jobs that had an error, not visible when n/a
* these 2 together, not visible when n/a
* redo all but replacing the original pattern with the explicit list of minions, al;ways there

### DIFF
--- a/saltgui/static/scripts/output/outputDocumentation.js
+++ b/saltgui/static/scripts/output/outputDocumentation.js
@@ -79,7 +79,7 @@ class OutputDocumentation {
     if(filterKey === "wheel") {
       return "";
     }
-    if (filterKey.startsWith("wheel.")) {
+    if(filterKey.startsWith("wheel.")) {
       // strip the prefix "wheel."
       return filterKey.substring(6);
     }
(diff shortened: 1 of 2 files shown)


erwindon:
@dawidmalina can you please review this one.
menu item titles may need improvement?

erwindon:
`Re-run job` uses the original target, which may be a wildcard. When you re-run a job, the list of minions may have changed compared to the original run. This menu item uses the new list.
`Re-run job (all)` uses the list of minions from the job. That does not automatically include new minions. This menu item uses the old list.

dawidmalina:
Ok 2nd test case (stop one of the minion) and run the same command:)

I would propose to change a bit description as currently are (imho) self descriptive. For example:
- Re-run job (on all minions)...
- Re-run job (on unsuccessful minions)...
- Re-run job (on failed minions)...
- Re-run job (on non responding minions)...

But on the other hand names are long :(

erwindon:
> But on the other hand names are long

used these texts, but without the parentheses (also saves 2 characters!)

PR #166, opened 2019-02-02
preview for selected grains

### LINKED ISSUE #161: allow to display custom grains

In our case we are configuring two quite important grains for each nodes (`environment` and `roles`).
First is just a regular string and 2nd is just a list of roles that minion have. It would be nice to allow (probably via master configuration) users to configure which additional grains should be display on grains page.

Alternatively it could be display on minions page as well.

**Additional context**

erwindon:
closes #161

### DIFF
--- a/saltgui/static/scripts/routes/grains.js
+++ b/saltgui/static/scripts/routes/grains.js
@@ -8,6 +8,28 @@ class GrainsRoute extends PageRoute {
 
     this._updateKeys = this._updateKeys.bind(this);
     this._updateMinion = this._updateMinion.bind(this);
+
+    // collect the list of displayed minions
+    let previewGrainsText = window.localStorage.getItem("preview_grains");
+    if(!previewGrainsText || previewGrainsText === "undefined") {
+      previewGrainsText = "[]";
+    }
+    this._previewGrains = JSON.parse(previewGrainsText);
+    if(!Array.isArray(this._previewGrains)) {
+      this._previewGrains = [ ];
+    }
+    // add all the required columns
+    const tr = document.querySelector("#page_grains thead tr");
+    while(tr.childElementCount > 6) {
+      tr.removeChild(tr.lastChild);
+    }
+    for(let i = 0; i < this._previewGrains.length; i++) {
+      const th = document.createElement("th");
+      th.innerText = this._previewGrains[i];
+      tr.appendChild(th);
+    }
+    // the new columns are not yet sortable, make sure they are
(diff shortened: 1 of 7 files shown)


saltgui/static/scripts/routes/grains.js
```diff
@@ -8,6 +8,28 @@ class GrainsRoute extends PageRoute {
...
+      this._previewGrains = [ ];
+    }
+    // add all the required columns
+    const tr = document.querySelector("#page_grains thead tr");
+    while(tr.childElementCount > 6) {
+      tr.removeChild(tr.lastChild);
+    }
+    for(let i = 0; i < this._previewGrains.length; i++) {
+      const th = document.createElement("th");
+      th.innerText = this._previewGrains[i];
```
dawidmalina: It would be nice to have first letter in uppercase to be consistent with rest of the columns names. It can be easily done by this:
```
th.innerText = this._capitalizeFirstLetter(this._previewGrains[i]);
```

where new helper function looks like this:
```
  _capitalizeFirstLetter(string) {
    return string.charAt(0).toUpperCase() + string.slice(1);
  }
```

erwindon: I had thought about that and decided not to do that. I like that they look slightly different. It visually also explains why they are on the right hand side. If anything must be done later, I would prefer an explicit column label value in the configuration.

dawidmalina: Yes, but please let try to think a bit about motivation of the user who would like to promote some grains (as a special one) to top level. I don't think he would like to see them behaves differently? If you will take my as example I would like to promote `roles` and `environment` grains as a first class citizens inside grains table. If you still not convinced can you please at least consider configurable option where I could configure it. Please, please, please :)

erwindon: not in this PR. and also unlikely later. additional reason is that this also does not happen on grains-details page.

erwindon:
@dawidmalina can you please verify whether this is good for merge?

dawidmalina:
Yes, we can merge it 👍 

Just one question. Why do we need: `white-space: pre-wrap;`? In my case when I'm using nested output proper list is generated but it's always wrapped even if I have big screen and a lot of space.
zrzut ekranu 2019-02-3 o 11 18 05

erwindon:
@dawidmalina can you check again please?

dawidmalina:
👍 ready to go :)

Issue #159, opened 2019-01-30
UX keys page improvments


dawidmalina:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
It would be more valuable to have keys related info on the keys page. For example we could remove: Salt version and OS version and instead we could display key fingerprint.

**Describe alternatives you've considered**

**Additional context**
We need to extend api to allow to new call `GET /keys/<minion id>` and / or introduce local cache.

dawidmalina:
@erwindon can you check my branch: `keys` it's almost working but I have issue with active minions which are not updated as td fingerprint is replaced by td os with content `loading...`. Can you take a look I believe you know how to fix this.

Currently it looks like this:
![zrzut ekranu 2019-01-31 o 23 34 08](https://user-images.githubusercontent.com/5910851/52090255-17c90580-25b1-11e9-89ae-98d3974fe494.png)

Issue #115, opened 2018-12-31
UX keys information page


dawidmalina:
I think that keys page can be more dedicated to keys then minion (in current version). If we could remove current content duplication: `Salt version`, `OS version` and introduce key fingerprint (https://docs.saltstack.com/en/latest/ref/wheel/all/salt.wheel.key.html#salt.wheel.key.finger). It can be simple fetched by adding new method in saltgui/static/scripts/api.js during keys iteration:
```
  getKey(id) {
    return this.apiRequest("GET", "/keys/" + id, {});
  }
```

To improve performance we could cache fingerprint of accepted minions and only fetch unaccepted keys.

erwindon:
I want to get rid of the _high-level_ api calls, and use only direct api calls.
In this case that would mean replacing `/keys` with `wheel.key.finger` with parameter `"*"` and still get all keys at once.
For rejected/denied/unaccepted keys no version information is shown (and cannot be shown). That leaves space to show the key.
Still not perfect.

erwindon:
superseded by #159

PR #167, opened 2019-02-02
ux keys page improvments

dawidmalina:
closes: https://github.com/erwindon/SaltGUI/issues/159
closes: #115

### DIFF
--- a/saltgui/static/scripts/routes/keys.js
+++ b/saltgui/static/scripts/routes/keys.js
@@ -36,21 +36,17 @@ class KeysRoute extends PageRoute {
       if(property === "local") continue;
       const hosts = keys[property];
       for(const hostname of Object.keys(hosts)) {
-        const fingerprint = hosts[hostname];
-        // why this is sometimes null?
-        const element = document.querySelector("#" + hostname);
-        if(element) {
+        const item = document.querySelector("#" + hostname + " .os");
+        if(item) {
           // remove td.os for accepted minions and add td.fingerprint
-          for (const item of element.children) {
-            if(item.className === "os") {
-              element.insertBefore(Route._createTd("fingerprint", ""), item);
-              element.removeChild(item);
-            }
-          }
-          // update td.fingerprint with fingerprint value
-          const fingerprintElement = document.querySelector("#" + hostname + " .fingerprint");
-          if(fingerprintElement) fingerprintElement.innerText = fingerprint;
+          item.parentElement.insertBefore(Route._createTd("fingerprint", ""), item);
+          item.parentElement.removeChild(item);
         }
+
+        // update td.fingerprint with fingerprint value
+        const fingerprintElement = document.querySelector("#" + hostname + " .fingerprint");
+        const fingerprint = hosts[hostname];
+        if(fingerprintElement) fingerprintElement.innerText = fingerprint;
       }
     }
   }
--- a/saltgui/static/scripts/routes/keys.js
+++ b/saltgui/static/scripts/routes/keys.js
@@ -6,23 +6,57 @@ class KeysRoute extends PageRoute {
     this.jobsLoaded = false;
 
     this._updateKeys = this._updateKeys.bind(this);
+    this._updateKeysFingerprint = this._updateKeysFingerprint.bind(this);
   }
 
   onShow() {
     const keys = this;
+    const p1 = keys.router.api.getKeys();
+    const p2 = keys.router.api.getKeysFingerprint();
+
+    Promise.all([p1, p2])
+      .then(function(data){
+        // process result of 1st promise
+        keys._updateKeys(data[0]);
+        // process result of 2nd promise
+        keys._updateKeysFingerprint(data[1]);
+      });
     return new Promise(function(resolve, reject) {
       keys.resolvePromise = resolve;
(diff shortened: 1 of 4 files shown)


dawidmalina:
@erwindon it's work in progress but I would like to have your quick feedback.

It's almost working but I have issue with active minions which are not updated as td fingerprint is replaced by td os with content `loading...`. Can you take a look I believe you know how to fix this.

Currently it looks like this:
zrzut ekranu 2019-01-31 o 23 34 08

erwindon:
The TD in the ACCEPTED rows still have id `os` instead of `fingerprint`.
For ONLINE minions, the row is created in `page.js`, function `_addMinion`.

saltgui/static/scripts/routes/keys.js
```diff
@@ -6,23 +6,44 @@ class KeysRoute extends PageRoute {
...
       keys.router.api.getJobsActive().then(keys._runningJobs);
     });
   }
 
+  _updateKeysFingerprint(data) {
+    const keys = data.return[0].data.return;
+    // console.log(keys);
+
+    for(const property of Object.keys(keys)) {
+      if(property !== "local") {
```
erwindon: to prevent a nesting level for exceptions
if(property === "local") continue;

erwindon:
Please also add `wheel.key.finger` to the list of exceptions in `page.js`

saltgui/static/scripts/routes/keys.js
```diff
@@ -6,23 +6,50 @@ class KeysRoute extends PageRoute {
...
+  _updateKeysFingerprint(data) {
+    const keys = data.return[0].data.return;
+
+    for(const property of Object.keys(keys)) {
+      if(property === "local") continue;
+      const hosts = keys[property];
+      for(const hostname of Object.keys(hosts)) {
+        const fingerprint = hosts[hostname];
+        // why this is sometimes null?
+        const element = document.querySelector("#" + hostname);
```
dawidmalina: @erwindon very often I have race condition when `keys.router.api.getKeysFingerprint()` is retrieved quicker then `keys.router.api.getKeys()` and then I don't have yet pre-filled informations so `_updateKeysFingerprint` will not work. Do you know what I can do to force proper order (some kind of chaining maybe)?

erwindon: You can combine multiple functions using `Promise.all` and use a single handler to process both results at the same time (calling the 2 original handlers sequentially). See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all.
This should be generalized for all API functions on all pages so that none of the additional details are processed before the page structure has been set up.

dawidmalina:
@erwindon `Promise.all` did the trick. Can you please take a look. It's first time I've been using it so let me know if you would do something differently.

erwindon:
Then you are ahead of me. I never used it! but it looks good, and I trust your tests.

dawidmalina:
👍 ready to go :)

saltgui/static/scripts/routes/keys.js
```diff
@@ -6,23 +6,57 @@ class KeysRoute extends PageRoute {
...
       keys.router.api.getJobs().then(keys._updateJobs);
       keys.router.api.getJobsActive().then(keys._runningJobs);
     });
   }
 
+  _updateKeysFingerprint(data) {
+    const keys = data.return[0].data.return;
+
+    for(const property of Object.keys(keys)) {
+      if(property === "local") continue;
```
erwindon: Just noticed... The answer has exactly 2 keys: `local` and `minions`. Just pick the `minions` part instead of looping through all available keys.

saltgui/static/scripts/routes/keys.js
```diff
@@ -6,23 +6,57 @@ class KeysRoute extends PageRoute {
...
+      const hosts = keys[property];
+      for(const hostname of Object.keys(hosts)) {
+        const fingerprint = hosts[hostname];
+        // why this is sometimes null?
+        const element = document.querySelector("#" + hostname);
+        if(element) {
+          // remove td.os for accepted minions and add td.fingerprint
+          for (const item of element.children) {
+            if(item.className === "os") {
```
erwindon: better just remove all content from the `tr`. then add all needed fields. that is what happens on other pages

dawidmalina: yes but you must loop again over different minion statuses to build proper drop down menu which is waist of cpu cycles. If it was simple menu entry then probably yes but here I don't think it's beneficial, better loop over small children and remove os.

saltgui/static/scripts/routes/keys.js
dawidmalina: it has more to be strict:
- local - master
- minion - accepted minions keys
- minions_rejected - rejected minions keys
- minions_pre - not accepted minions keys

saltgui/static/scripts/routes/keys.js
dawidmalina: @erwindon is this explanation enough?

dawidmalina:
Anything else I should modify here before we can merge it?

erwindon:
I'll continue on this branch and make a few changes before the actual merge.

PR #177, opened 2019-02-10
show keys with monospace font

erwindon:
Currently, keys are shown in a proportional font on the Keys page.
This looks slightly chaotic.
This PR sets a monospace font only for that column.

### DIFF
--- a/saltgui/static/stylesheets/keys.css
+++ b/saltgui/static/stylesheets/keys.css
@@ -1,0 +1,3 @@
+.fingerprint {
+  font-family: monospace;
+}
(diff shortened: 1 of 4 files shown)


saltgui/static/stylesheets/keys.css
```diff
@@ -0,0 +1,3 @@
+.fingerprint {
+  font-family: monospace;
```
dawidmalina: Looks good. One think (at least in my browser) text is vertically align to the top. It looks much better if we will add `vertical-align: middle;`. What do you think.

Btw this I believe is common issue:
zrzut ekranu 2019-02-10 o 15 55 25

erwindon: Same here. The solution affects the placement of some elements in situation with higher rows. Such as when viewing grain values.

erwindon:
Can you look at it again?

dawidmalina:
LGTM 👍

PR #173, opened 2019-02-04
closing job-details page should redirect to previous page

dawidmalina:
closes: https://github.com/erwindon/SaltGUI/issues/139

PR #186, opened 2019-03-03
always use the regular api

### LINKED ISSUE #182: solve name conflict between API and pages
Currently the URLs for the pages of SaltGUI share the namespace with the API calls.
e.g. `/keys` is a page when `Accept: text/html` is used, but it is and API call when `Accept: application/json` is used. This is confusing.

Use a separate prefix for pages. e.g. `/page/`.

erwindon:
for #182 and other reasons, the api functions must be sanitized:
* only use the direct api functions (prevents name conflicts)
* rename all api functions according to the api function that they are calling
* organize the api functions alphabetically

closes #182

### DIFF
--- a/saltgui/static/scripts/api.js
+++ b/saltgui/static/scripts/api.js
@@ -51,42 +51,22 @@ class API {
       });
   }
 
-  getMinions() {
-    return this.apiRequest("GET", "/minions", {});
-  }
-
-  getKeys() {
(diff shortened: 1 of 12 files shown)


erwindon:
@dawidmalina can you please review/test this one? should be a good basis for #173

dawidmalina:
Looks Good :+1:

PR #216, opened 2019-04-21
Use jobs.active to show which output is still expected

### LINKED ISSUE #206: Use jobs.active to show which output is still expected
When job details are shown, all minions without data are treated equally. It is assumed that the job has timed out. This a result of the usage of call `jobs.list_job`.
But the jobs overview(s) already show how many minions are still working on a job.

Use function `jobs.active` provides the difference between minions that timed out and minions that are still working on the request.

erwindon:
closes #206

When job details are shown, all minions without data are treated equally. It is assumed that the job has timed out. This a result of the usage of call jobs.list_job.
But the jobs overview(s) already show how many minions are still working on a job.

Use function jobs.active provides the difference between minions that timed out and minions that are still working on the request.

### DIFF
--- a/saltgui/static/scripts/output/Output.js
+++ b/saltgui/static/scripts/output/Output.js
@@ -212,6 +212,7 @@ export class Output {
       // runners/wheel responses are not per minion
       // Do not produce a #response line for async-start confirmation
       const span = document.createElement("span");
+      span.id = "summary";
 
       const cntResponses = Object.keys(response).length;
       const cntMinions = minions.length;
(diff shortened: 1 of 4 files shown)


erwindon:
Initial screen (because retrieving the status is slower that retrieving the job details):
afbeelding

Some minions are still working on the job, note that "(no response)" is replaced by "(active)"; and in a distinctive and more positive colour:
afbeelding

All minions are done:
afbeelding

erwindon:
@dawidmalina can you please take a quick look at this one, and see whether it also works for you?
thx!

PR #218, opened 2019-04-24
allow jobs to be stopped, either as job, or per minion

erwindon:
The job overview (since recently) shows which minions are still working on a job.
It should be possible to:
* stop the job on all minions
* stop the job per minion
* show information on a job per minion

This PR adds:
* menu item on job level to stop the job. the menu item is hidden again when it turns out that the job is already completed
* a link on the output panel to allow the job to be stopped on that minion
* a link on the output panel to allow information on the the job to be shown for that minion


dawidmalina:
Nice improvment :+1:

erwindon:
any chance that you can do a small try?
e.g. "test.rand_sleep 300" on all minions and then click around?

dawidmalina:
Looks ok:) Would be good if page could be reloaded (refreshed) after job was killed. Currently when I closed manual run box nothing change - I need to reload page - it can be misleading.

dawidmalina:
And one think - "terminate job" menu could be also added to recent jobs section to allow job termination.

erwindon:
> Looks ok:) Would be good if page could be reloaded (refreshed) after job was killed. Currently when I closed manual run box nothing change - I need to reload page - it can be misleading.

I'll look into that...

erwindon:
> And one think - "terminate job" menu could be also added to recent jobs section to allow job termination.

Actually there are 3 commands available: `saltutil.kill_job`, `saltutil.term_job` and `saltutil.signal_job`. I chose for `signal_job` with explicit parameter `15` (or `terminate`). User can change that to any signal, e.g. `9` (or `kill`), before starting the command.
But that may be too complicated for non-aware users.
I'll just add all 3 commands.

erwindon:
This one is ready.
Do you like to give it a quick test-run before I merge it?

dawidmalina:
Looks good. We have now full options to handle jobs - but I must agree with you it can be complicated for regular users (especially what signal id should be used) :|

erwindon:
One can always click the documentation icon for immediate first explanation...

PR #220, opened 2019-04-27
Rework on promise creations/executions

### LINKED ISSUE #195: Invalid promise handling
**Describe the bug**
Several problems in Promise handling

**To Reproduce**
Are not needed

**Expected behavior**
Promises should also tackle the rejection case. This is mostly not given. 

**Example**
```javascript
if(route.onShow) response = route.onShow();
```

It is never checked if inner promises resolve or not.

**Additional context**
reason could be found in bad test design.

if I will find time after the Webpack PRs i can tackle this issues.

erwindon:
closes #195 

Problems:
* code involving promises looks bad
* code involving promises uses its own administration (e.g. `keysLoaded`)
* promises are used in multiple layers
* sequence of promises is not guaranteed

Work done:
* Once created, each promise is first created as a constant with a standardized name
* no extra promise layer is used
* sequence of promises is now managed
* extra administration is removed
* added error handling for technical errors from normal salt commands
* added error handling for technical errors from RUNNER salt commands
* added error handling for technical errors from WHEEL salt commands

TODO:
* only allow tooltips from SPANs, not from TDs

### DIFF
--- a/docs/PERMISSIONS.md
+++ b/docs/PERMISSIONS.md
@@ -47,6 +47,7 @@ The following configuration is a mimimum set of permissions, so that SaltGUI can
         - jobs.list_jobs
       - '@wheel':
         - config.values
+        - key.finger
         - key.list_all
 ```
 Adititional permissions are needed to run the commands associated with the popupmenu items.
--- a/saltgui/static/scripts/Api.js
+++ b/saltgui/static/scripts/Api.js
@@ -63,7 +63,7 @@ export class API {
       params.tgt_type = "glob";
       params.tgt = "*";
     }
-    return this.apiRequest("POST", "/", params).catch(console.error);
+    return this.apiRequest("POST", "/", params);
   }
 
   getLocalPillarItems(minion) {
@@ -78,7 +78,7 @@ export class API {
       params.tgt_type = "glob";
       params.tgt = "*";
     }
-    return this.apiRequest("POST", "/", params).catch(console.error);
+    return this.apiRequest("POST", "/", params);
   }
 
   getLocalPillarObfuscate(minion) {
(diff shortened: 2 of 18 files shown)


erwindon:
@smarletta This PR improves several aspects of the "promise" handling of SaltGUI, see the description of thus PR. Do you see anything that still should be done before #195 is considered done?

@dawidmalina fyi

erwindon:
@dawidmalina 
This PR seems stable now...
Whenever an API error occurs, the error message is now added to the GUI.
Do you have any comment on the design or the implementation?

The best way to force an api error is to remove some categories from {{external_auth}} in the file {{/etc/salt/master}}; or to specify the only allowed api calls for a category. In both cases also restart salt system of course.
But you can also break the code in {{Api.js}} by specifying an unknown {{client}} or {{fun}} in one or more of the helper functions.

I'll still do some code cleanup, mainly to reduce some code duplications.

dawidmalina:
This a lot of code cleanup and refactoring and I like it:) Nice work :+1:

erwindon:
Ok, thx!
Will you be able to do a quick test in the coming days?

dawidmalina:
Yes. But next week.

erwindon:
this is worth waiting for that

dawidmalina:
@erwindon Can you confirm that template page is not working for you as well? Looks like wheel config.values does not fetch any configuration.

dawidmalina:
Would be nice if only text background will be highlighted not all td. 
Screenshot 2019-05-06 at 21 34 00

Can we change this to (for example):
```
<td class="fingerprint"><span class="tooltip">(error)<span class="tooltiptext">{"status":400,"message":"Bad Request"}</span></span></td>
```

dawidmalina:
Rest looks good and I could found any issues or regressions. Good job 👍

erwindon:
thx!
> Would be nice if only text background will be highlighted not all td.

The existing elements were re-used, without affecting the structure.
Since I was already in doubt about this one, I'm now working on this.

PR #221, opened 2019-05-03
add table search function

erwindon:
Depending on the environment, the pages may get a bit long.
A search box is added. The table will then only show the rows that somehow contain this text.

e.g.:
afbeelding

Done:
* add facility to hide/show the search boxes. initially not shown.
* fix search facility on pillar-details screen, which has some problems
* added facility to highlight the found text
* are the job list first (A) "length restricted THEN searchable" or should they be (B) "searchable and then length restricted". Currently (A) applies. (B) may be more useful. ==> the current situation (A) remains. We probably will do (B) later.

### DIFF
--- a/saltgui/index.html
+++ b/saltgui/index.html
@@ -41,6 +41,7 @@
       <div class='fab' id='button_manualrun'>&gt;_</div>
       <div class="minimenu">
         <div class="dropdown">
+          <!-- 2261 = MATHEMATICAL OPERATOR IDENTICAL TO (aka "hamburger") -->
           <div class="menu_item">&#x2261;</div>
           <div class="dropdown-content">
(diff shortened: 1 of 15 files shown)


dawidmalina:
This is nice improvements 👍 Just wandering whether recent jobs (box on the right side) shall be searchable? We are always returning few (7 if i remember correctly).

And for the regular job page I think that we should search through filtered jobs only. If we will first search then filter against result length then we might return job which will be never visible with regular view. So 1st option would be more natural unless we will fetch all jobs then 2nd would be more suitable.

erwindon:
@dawidmalina @ellipsion
How do you like the table search function as it is in this PR?

dawidmalina:
For me looks ok.

PR #254, opened 2019-07-06
Timeout

### LINKED ISSUE #251: After errors due to session-timeout, SaltGUI should return to login screen
When there is a session timeout, most API calls end with "401 Unauthorized".
SaltGUI carries on because it may not have access to a single API call (or range of API calls).

When a 401 is received, the current time should be compared with the session-end time as received at login. When the current time is past that, it is safe to conclude that the whole session has expired. The logout procedure should be triggered so that cached information is removed and the screen reverts to the login screen.
Function `isAuthenticated` can be retired.

erwindon:
When there is a session timeout, most API calls end with "401 Unauthorized".
SaltGUI carries on because it may not have access to a single API call (or range of API calls).

When a 401 is received, the current time should be compared with the session-end time as received at login. When the current time is past that, it is safe to conclude that the whole session has expired. The logout procedure should be triggered so that cached information is removed and the screen reverts to the login screen.
Function `isAuthenticated` can be retired.

closes #251

### DIFF
--- a/saltgui/index.html
+++ b/saltgui/index.html
@@ -66,7 +66,7 @@
       <div class='route' id='page-login'>
         <form method='post' id='login-form'>
           <h1>SaltGUI</h1>
-          <div class='notice-wrapper'>
+          <div id='notice-wrapper'>
           </div>
           <input id='username' type='text' placeholder='Username' autofocus />
           <input id='password' type='password' placeholder='Password' />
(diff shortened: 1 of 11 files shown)


erwindon:
Note that this branch is a sub-branch of `options` (so it cannot be merged now).

erwindon:
@lostsnow  @dawidmalina @elipsion
Do you have any experience with `nightmare`?
(i.e. the functional regression-test that runs as part of `runtests.sh`)
The regression test now fails and I cannot find how to get the proper error message...

lostsnow:
Try this: https://github.com/mochajs/mocha/issues/2025?

PR #257, opened 2019-07-14
actions for jobs now in dropdown

### LINKED ISSUE #219: Links for process management should be in a dropdown menu
When viewing job, 4 links are shown for each running process within that job.
The titles are `info`, `term`, `kill` and `signal`

These links should be organized in a drop-down menu.
The titles should be `Show process info`, `Terminate process`, `Kill process` and `Signal process`

**Additional context**
The original implementation had only one link, then it became two, then four.
Now it is time to clean-up.

erwindon:
closes #219 

When viewing job, 4 links are shown for each running process within that job.
The titles are `info`, `term`, `kill` and `signal`

These links are now organized in a drop-down menu.
The titles should be `Show process info`, `Terminate process`, `Kill process` and `Signal process`

### DIFF
--- a/saltgui/static/scripts/DropDown.js
+++ b/saltgui/static/scripts/DropDown.js
@@ -18,21 +18,21 @@ export class DropDownMenu {
 
     this.menuDropdown = Route.createDiv("run-command-button", "");
 
-    switch (pParentElement.id) {
-    case "cmd-box":
+    if(pParentElement.id === "cmd-box") {
       // 1F4D6 (D83D+DCD6) = A BOOK
       this.menuButton = Route.createDiv("menu-dropdown", "\uD83D\uDCD6");
-      // hide the menu until it receives menu-items
-      this.verifyAll();
-      break;
-
-    default:
+    } else if(pParentElement.classList && pParentElement.classList.contains("minion-output")) {
+      // 1F4D6 (D83D+DCD6) = A BOOK
+      this.menuButton = Route._createSpan("menu-dropdown", "\u2261");
+    } else {
       // 2261 = MATHEMATICAL OPERATOR IDENTICAL TO (aka "hamburger")
       // assume it will be a command menu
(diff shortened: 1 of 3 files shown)


erwindon:
@dawidmalina 
Dawid,
You thumbed-up the original issue #219. Can you have a quick look at it? It needed a few counter-measures in the CSS area.
But I'm also interested whether I missed something in the latest release (this PR is build on top of that). Specifically the session management with the new forced logout after session-timeout. Do you have a few minutes for that too?
thx!

erwindon:
btw: still working on the functional tests that fail since the last release...

dawidmalina:
Sure. I will try to validate this in the coming days.

dawidmalina:
@erwindon I've checked drop down menu and it's working ok 👍 

According session timeout after few hours session was terminated and expired. BTW do you know how to tune this timeout? It should (in my opinion) counted in minutes not hours.

There are two parameters related to session:
- session_id
- token
do you know how to configure it's lifetime?

Session_id removal (cookie) not triggering session end but in my opinion should. I'm not security expert but would be good to validate session management.

erwindon:
> [session timeout]
> BTW do you know how to tune this timeout?
> It should (in my opinion) counted in minutes not hours.
Parameter `token_expire` in file `/etc/salt/master` controls the session-timeout.
Its value is in seconds. It seems to be counted from the login call. This also means that the session can expire while one is actively using the salt-api, e.g. from SaltGUI.

erwindon:
> There are two parameters related to session: session_id and token
> do you know how to configure it's lifetime?

These parameters are both removed by the logout procedure.

> Session_id removal (cookie) not triggering session end but in my opinion should.
> I'm not security expert but would be good to validate session management.

salt-api can be used with either `session-id` or `token`.
Both are set on login.
SaltGUI uses only `token`.
Therefore the removal of `session-id` is not relevant.

erwindon:
thx!

Issue #260, opened 2019-07-20
functional regression tests fail


erwindon:
**Describe the bug**
The login/logout code was refactored.
After that, the functional regression tests fail.
See screenshot.

**To Reproduce**
See this GitHub project.

**Expected behaviour**
Functional regression run successfully.
Or at least does not fail due to non-related technical issues.

**Screenshots**
![afbeelding](https://user-images.githubusercontent.com/3663742/61583139-3a85b100-ab34-11e9-969e-ff4c103bd7da.png)

**help wanted**
My knowledge about the `nightmare` package is limited. e.g. I'm not able to see what the actual error is. And that makes it hard to fix a problem. The best type of help is an explanation how to see the actual error. The next best type of help is an explanation is an update of the tests to get the tests successful again.

erwindon:
since the tests sometimes succeed on GitHub, I think that there are no real steps to be taken here.
we'll leave this for now.
obviously, any improvement suggestion on the functional tests is appreciated...

PR #264, opened 2019-10-19
functional regression tests fail #260

dawidmalina:


dawidmalina:
Hmm this is strange. Tests are passing in my local environment:

```
  Funtional tests
    Login and logout
      ✓ we should be redirected to the login page (2375ms)
      ✓ we cannot login with false credentials (6287ms)
      ✓ valid credentials will redirect us to the homepage and hide the loginform (2683ms)
      ✓ check that we can logout (3868ms)


  4 passing (15s)
```

The easiest way to debug is to extend the timeout and use:
```
NIGHTMARE_DEBUG=1 ./runtests.sh
```
and see real output.

Basically:
1. missing param `?reason=no-session`
2. new notice box `no-session` - need to wait a bit
3. redirection is taking some time - need to wait a bit
4. missing param `?reason=logout`

In addition 2, 3, 4 have strange issue where username is not provided with single `.type()` - I don't know why but at least (in my case) tests are passing.

erwindon:
I finally found the problem. More debugging was needed for that, also variable `DEBUG=nightmare:*,electron:*` was useful. Travis was missing library `libgconf-2-4` and could not run the tests.
fyi: now using `insert()` instead of `type()`.
I'm ignoring this PR, but I used all/most of it. I'll still add the credit in the notes.
thx!

Issue #268, opened 2019-10-31
Apply coloured output to output from runners.state.orchestrate


erwindon:
**Is your feature request related to a problem? Please describe.**
Output from `runners.state.orchestrate` is not recognized as highstate output within a `data` tag

**Describe the solution you'd like**
Hide the `data` tag and show the remaining output as highstate data

**Describe alternatives you've considered**
Show the raw data

erwindon:
Sample data:
```
    "data": {
        "devops-sm_master": {
            "salt_|-bad_state_|-bad_state_|-state": {
                "__id__": "bad_state",
                "__jid__": "20191031014909428714",
                "__run_num__": 1,
                "__sls__": "xxx/xyz",
                "changes": {
                    "out": "highstate",
                    "ret": {
                        "dev-erwindon-it-4090": [ "No matching sls found for 'fail_with_changes' in env 'base'" ]
                    }
                },
                "comment": "Run failed on minions: dev-erwindon-it-4090",
                "duration": 817.428,
                "name": "bad_state",
                "result": false,
                "start_time": "01:49:09.355361"
            },
            "salt_|-good_state_|-good_state_|-state": {
                "__id__": "good_state",
                "__jid__": "20191031014908597424",
                "__run_num__": 0,
                "__sls__": "xxx/xyz",
                "changes": {
                    "out": "highstate",
                    "ret": {
                        "dev-erwindon-it-4090": [ "No matching sls found for 'succeed_with_changes' in env 'base'" ]
                    }
                },
                "comment": "Run failed on minions: dev-erwindon-it-4090",
                "duration": 831.248,
                "name": "good_state",
                "result": false,
                "start_time": "01:49:08.523899"
            },
            "salt_|-mymod.myfunc_false_result_|-mymod.myfunc_false_result_|-function": {
                "__id__": "mymod.myfunc_false_result",
                "__jid__": "20191031014911731240",
                "__run_num__": 3,
                "__sls__": "xxx/xyz",
                "changes": {
                    "out": "highstate",
                    "ret": { "dev-erwindon-it-4090": "'mymod.myfunc_false_result' is not available." }
                },
                "comment": "Running function mymod.myfunc_false_result failed on minions: dev-erwindon-it-4090 Function mymod.myfunc_false_result ran on dev-erwindon-it-4090.",
                "duration": 1498.479,
                "name": "mymod.myfunc_false_result",
                "result": false,
                "start_time": "01:49:11.664238"
            },
[...]

PR #269, opened 2019-10-31
added support for orchestration output

erwindon:
See #268

e.g.:
afbeelding

### DIFF
--- a/saltgui/static/scripts/output/Output.js
+++ b/saltgui/static/scripts/output/Output.js
@@ -422,6 +422,12 @@ export class Output {
 
     let nrMultiLineBlocks = 0;
 
+    // convert state.orchestrate output back to regular highstate
+    if(pResponse.RUNNER && pResponse.RUNNER.outputter === "highstate") {
+      pResponse = pResponse.RUNNER.data;
+      pMinionData = Object.keys(pResponse);
+    }
+
     // for all other types we consider the output per minion
     // this is more generic and it simplifies the handlers
     for(const minionId of pMinionData.sort()) {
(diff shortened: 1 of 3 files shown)


erwindon:
@gnouts
I've added a bit of code to recognize the orchestration output and present it in the highstate format.
Can you please try it out in a more real-life situation? thx!

erwindon:
@gnouts
Do you have the opportunity to try it out in a more real-life situation? thx!

gnouts:
Yep I tried this branch with my orchestrators. It works, just like on your screenshot, that's nice :) 
Though I noticed it doesn't honor the state_output settings of the master. So the `ret` was really verbose.

erwindon:
OK, great!

> Though I noticed it doesn't honor the state_output settings of the master.
> So the ret was really verbose.

would you be able to provide an example of that data?
e.g. with `salt-run --out=json state.orchestrate ...`
When you mail it, it will not become visible for others, when that is an issue.

erwindon:
@gnouts 
I accidently skipped your remark about `state_output`.
SaltGUI does not use the variables like `state_verbose` or `state_output`.
But I made a request to support them in #270.

Was the output in your trial-run otherwise ok?
Would you be a able to provide an actual sample output?
(or better, the data behind the output)

gnouts:
@erwindon I'll try to provide you a test example but as I'm doing this in a profesionnal context I might not be able to do it before at least 2 weeks, sorry.
(You might want to merge this one and I'll open an issue as it's not really a major issue.)
Thanks for your work :)

erwindon:
OK!
the merge will happen on the next rainy day (i.e. probably next Saturday)

erwindon:
Sorry for the delay, last Saturday was indeed rainy, but too much other stuff to do...

Issue #334, opened 2020-11-27
migrate to use a proper web-application framework


erwindon:
**Is your feature request related to a problem? Please describe.**
The current version of SaltGUI uses its own simple navigation framework. While it works reasonably well, it has a long list of minor problems:
* it is a lot of extra code to maintain
* the back-button does not always work properly
* page refresh does not always work properly
* the page url is not always correct with the content that is shown
* hidden pages still contribute to the scrollbar length

some of these have their own issue (e.g. #229, #293, #327) but may benefit for a more solid framework

**Describe the solution you'd like**
select a web-application framework with the following properties:
* it must be client-only
* must be fully open/free
* must support a full application, from navigation 
* is popular and/or easy to use
* can still use most of the current styling and flow

The table below lists 9 frameworks as found on https://opensource.com/article/20/5/open-source-javascript-frameworks:
1. ReactJS
2. Angular
3. VueJS
4. MeteorJS
5. KnockoutJS
6. EmberJS
7. BackboneJS
8. Svelte
9. AureliaJS

Q: What can YOU do?
A: Point out the ones from this list that are not suitable for whatever reason, so that in the end we have a best choice!

bbinet:
So you would like to rewrite everything from scratch using one of these js frameworks? Sounds like quite a bit of work! ;-)
If this is just an issue with the router, why not try to only pick a router component like the following:
https://github.com/ChrisRu/es6-router
(I don't have any experience with it, I've just searched for a pure js router)

erwindon:
The router has the most actual problems.
But the gui itself has too much code, mostly in the lower layers of html object manipulation. so improvement there is welcome too.

part of the investigation would have been on how many changes are needed in the existing system.

saltgui already has a single page navigation facility. but I broke it 2 years ago with the first new feature (the keys page) that I supplied. I never realized that single page navigation was in place as saltgui at that time did not have additional pages anyway. and all this time we had this router doing basically nothing. this continued with all new pages. decision was between fixing and replacing (this issue). I was also investigating to fix the current router and it seems not too difficult. So I'll do that in parallel.

erwindon:
The tests with the refurbished (current) router are very promising.
We might not need to push the software into a standard framework.

bbinet:
That's great! 👏

erwindon:
#339 is a better direction

Issue #327, opened 2020-11-19
fix navigation issues


erwindon:
**Describe the bug**
...

**To Reproduce**
e.g. login, minions, minion, close minion --> logout
e.g. login, start cmdbox, close cmdbox --> logout

**Expected behaviour**
return to previous screen

**Screenshots**
...

**Additional context**
SaltGUI uses 2 navigation techniques:
1) traditional; url history is automatically done by the browser
2) the internal router; url history is maintained by the router

these 2 can deviate.
when using (2) only, the navigation is also smooth, using transitions between pages

complicating factor is that history administration (window.history) is barely debuggable

Issue #293, opened 2020-08-01
Closing minion-screen does not update the URL


erwindon:
**Describe the bug**
When navigating from grains-minion back to minions screen, the screen is properly updated, but the URL remains.
When a refresh is used, the screen reverts to its original state.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to 'Grains'
2. Go to any specific minion
3. See url ".../grains-minion?minionid=xxxx"
4. Click on 'X' to return to main grains screen
5. See that URL remains on ".../grains-minion?minionid=xxxx" instead of changing to ".../grains"

**Expected behaviour**
The url should change to ".../grains"

**Additional context**
This also happens for Schedules, Pillars and Beacons in the same way.
But it does not happen for the Job screen (which returns properly to the previous screen)

erwindon:
The job screen uses `window.history.back()` because it is otherwise unknown where we came from. i.e. that might have been from any of the 'jobs' side-panels, or from the 'jobs' page.
Solution is to use `window.history.back()` for all cases.

Issue #229, opened 2019-05-25
Jobs page shows scrollbar even with few jobs


erwindon:
**Describe the bug**
Even with minimal content on the screen, the page-scrollbar is still shown.

**To Reproduce**
1. Go to 'Jobs'
2. Show the filter box
3. Type "lkljhlk" (or any non-existant job name)

**Expected behavior**
Scrollbar is not visible when there is (almost) no content

**Additional context**
Cause is that pages are not truly hidden. They get "opacity:0".
This is due to the use of (subtle) animations, as animations do not work for "display:none".

erwindon:
Strange, this is actually explicitly handled in `hideRoute`...

erwindon:
working on choice nr 2

PR #339, opened 2020-12-08
Singlepage

erwindon:
This PR turns SaltGUI into a single-page web application.
Navigation is no longer between web-pages, but between states.
Don't worry, the states act as pages.

benefits:
* navigation is faster
* navigation has nice page transitions
* the back-button should now always work properly
* there is only one type of navigation

This may also solve one or more of #229, #293, #327.
This obsoletes #334.

### DIFF
--- a/saltgui/static/scripts/Api.js
+++ b/saltgui/static/scripts/Api.js
@@ -265,10 +265,10 @@ export class API {
           const loginResponseStr = Utils.getStorageItem("session", "login-response");
           if (!loginResponseStr) {
             this.logout().then(() => {
-              window.location.replace(config.NAV_URL + "/login?reason=no-session");
+              this.router.goTo("login", {"reason": "no-session"});
               return true;
             }, () => {
-              window.location.replace(config.NAV_URL + "/login?reason=no-session");
+              this.router.goTo("login", {"reason": "no-session"});
               return false;
             });
           }
@@ -280,10 +280,10 @@ export class API {
             const expireValue = loginResponse.expire;
             if (now > expireValue) {
               this.logout().then(() => {
-                window.location.replace(config.NAV_URL + "/login?reason=expired-session");
+                this.router.goTo("login", {"reason": "expired-session"});
                 return true;
               }, () => {
-                window.location.replace(config.NAV_URL + "/login?reason=expired-session");
+                this.router.goTo("login", {"reason": "expired-session"});
                 return false;
               });
             }
@@ -302,6 +302,12 @@ export class API {
     if (!tokenOnSetup) {
       return;
     }
+
+    // allow only one event-stream
+    if (API.eventsOK) {
+      return;
+    }
+    API.eventsOK = true;
 
     let source;
     try {
--- a/saltgui/static/scripts/DropDown.js
+++ b/saltgui/static/scripts/DropDown.js
@@ -139,4 +139,9 @@ export class DropDownMenu {
     this.menuDropdown.style.display = "none";
   }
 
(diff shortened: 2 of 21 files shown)


erwindon:
@bbinet can you please give this a quick/medium/full(your choice) test?

bbinet:
Good job @erwindon, this is great!
The only small issue I have seen is that if I open a jobid then close it by clicking on the up-right cross, the page is correctly updated, but not the url, so if I reload the page, the jobid page is shown again.

erwindon:
@bbinet not that I doubted your observation, but I see the same thing. And it is consistent with the other close-buttons, e.g. when viewing the per-minion details of grains/schedules/pillars/beacons.
The handling of the close-buttons is different than other navigation because the effect needs to be exactly the same as the back-button from the browser.

erwindon:
And I spotted another one. adding menu items to panel-level menu's was broken. since we are now re-using panels, the menus were also re-used. but in some cased the same menu items were being added again. I chose too quickly to erase all these menu's when navigating. the 3 pages (jobs/keys/minions) that added their menu item only once in the constructor lost them very quickly. I now need to decide between the 2 solutions instead a part of each...

erwindon:
fixed the panel-level dropdown menu's, these now always assume the world has changed and will determine again-and-again the visibility of their menu-items.
fixed the regular navigation for the 3 use-cases: (1) forward navigation, (2) close-button in application and (3) browser back-button.

Unfortunately, one thing remains... when a page is refreshed using the browser, the "previous-page" administration becomes overwritten with a reference to the same page, which is actually the truth. no escape possible...

erwindon:
@bbinet
I think I've addresses all the issues we found.
can you give it another try?

erwindon:
@bbinet can you please let me know whether you can try this again?

erwindon:
@dawidmalina you previously helped with the graphical design of SaltGUI. I was now finally able to make it into a Single Page Application. Can I invite you to try it out from this branch and let me know what you think?

bbinet:
@erwindon Sorry for the delay, Christmas period is very busy...
I've just tested your updated singlepage pull request again, and everything seem perfect, good job!

dawidmalina:
@erwindon good job. I did a few basic tests - same as i did in the past and all use cases are properly handled.

I've found few topics:
- pause icon is always present on the keys page
- pause icon is always present on the show beacons page
- run manual icon is (at least on my browser) rather elliptic

Not sure are those issues or not.

Hope i could help.

erwindon:
@dawidmalina 
> * pause icon is always present on the keys page
> * pause icon is always present on the show beacons page
Both the keys page and the beacons page get asynchronous updates from the salt event stream.
The pause button is a way to pause that mechanism.
(the button is also present on the events page for the same reason; and on the job-details page to stop collecting status details for each and every visible job during initial screen population or after scrolling new jobs into view)
does this explanation cover the observation?

> * run manual icon is (at least on my browser) rather elliptic
I'm quite sure nothing in this branch changed that. so likely it will be treated as a separate issue.
fyi: that 'icon' is actually the text `>_` with a big rounded green border. so technically there is no instruction present to make it a circle. but we all assume that is should be.
can you please raise this as a separate issue and including a screen-shot of that area?
very likely, I will then turn it into an svg
(It took me just 10 minutes to do that now in my workspace)

dawidmalina (approved):

erwindon:
@dawidmalina
thx!
I'll push out a version+release with what was committed before this.
this will then be merged into the master branch for the next version/release.

PR #344, opened 2020-12-25
make the cmd-run button a true circle

erwindon:
Problem:
The current run-button is built from the text `>_`, which is given an html-border.
That border is relatively thick and seems to form a circle.
However, it is not a circle as it is just a border around `>_`.
It looks like:
before

Solution:
Style the run-button as SVG
It now looks like:
after

### DIFF
--- a/saltgui/index.html
+++ b/saltgui/index.html
@@ -50,9 +50,9 @@
         <div class='menu-item' id='button-logout1'>logout</div>
       </div>
 <!-- mini menu -->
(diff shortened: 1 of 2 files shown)


dawidmalina (approved):

erwindon:
@dawidmalina 
thx!
I assume you saw the same original shape?

dawidmalina:
Yes. It is looking good with this simple change. I like it 👍

PR #299, opened 2020-12-27
Update the code to better handle the dropdown menus

erwindon:
The current "checkbox-button" style menu and the "radio-button" style menu's have very dirty implementations.
Both can be found in the command-box.

This should be cleaned up. Best solution seems to be the introduction of a "DropDownCheckbox" and "DropDownRadio" sub-classes that handle all the specifics.

1) Leave it like this.
2) switch to a gui toolkit that has this available.

**Additional context**
Additionally, the difference between the user's choice and the system's choice in the "target" menu should be enhanced.

Additionally, the dropdown menu's should work on touchscreens. i.e. the menu opens on 'click' instead of only 'mouseover'.


erwindon:
build completed
TODO: simplify _updateTableFilter to remove sonarcloud issue
TODO: more testing

PR #329, opened 2020-12-27
Add support for creating templates from SaltGUI directly

erwindon:
It would be great if we could add a button "Save as template" from the command box dialog so that we could store the current command as a template in the local storage of the browser, so that the command can be easily replayed later on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,8 @@ Note that the effect is achieved by string truncation only. This is equivalent t
 ## Templates
 SaltGUI supports command templates for easier command entry into the command-box.
 The menu item for that becomes visible there when you define one or more templates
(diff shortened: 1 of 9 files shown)


erwindon:
this one has a bit of a delay.
the reason is mainly that this change became too complex/extended/complicated for its functionality.
the functionality included applying templates from different locations and saving/deleting local templates.
for saving/deleting local templates, an internal command system was created.

I will now simplify this change a bit by not adding the save/delete capabilities.
But the templates can still be used from definitions in multiple locations.

the branch will be renamed to `use_templates`

erwindon:
this change became too complex to handle.
I'll continue to allow reading templates from other locations, but not changing them

PR #330, opened 2020-12-27
Add support for configuring which pages are displayed on SaltGUI

erwindon:
I would like to allow some people in our team to access SaltGUI to run salt commands on the minions they are allowed to control, and it would be great to be able to hide the tabs they won't use.

The best solution would be to have a way to configure which pages a user is allowed to see (maybe through extra configuration in the salt-master config like this is currently done for the templates feature), then based on this user config, SaltGUI could hide the pages that the user is not allowed to see.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ saltgui_pages:
     - keys
     - grains
 ```
-Note that this is NOT as security mechanism to reduce what a user can do.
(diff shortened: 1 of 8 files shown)


erwindon:
@bbinet Bruno,
I've finally merged the `singlepage` branch which indirectly solves the problems you reported.
A few corrections were needed to match the new reality of that merge.
Can you please verify this change again?

PR #338, opened 2020-12-27
improve special character handling

erwindon:
SaltGUI relies on the support for about 20 non-ascii characters in the browser.
These are identified by unicode code-point.
In some environment the fonts for these characters are not available.
As an additional problem, some of the characters are represented as "emoji" style characters. i.e. using colours, even though the instruction to suppress colours is present.

An alternative that always works correctly.
Last remaining alternative:
* supply a (free!) web-font, which includes all these characters, use for all text.

In all cases, the handling must be delegated to a single class.

* leave as is, but provide documentation instructions to make the proper fonts available.
no-one reads the documentation, so that does not solve anything.
* supply a (free!) web-font, which includes all these characters, use only for special characters.
switching back/forth between fonts becomes very ugly.
we then cannot use assignments to `innerText` any more.
* supply a (free!) image set, which includes all the these characters as images
lots of problems on resizing the images.
and again, we then cannot use assignments to `innerText` any more.

**Additional context**
There are many environments that may each have their own problems.


erwindon:
In the recent Debian 11, the included firefox no longer has this presentation problem.
closing this issues until someone points to a sensible OS-version + BROWSER-version that still has problems.
the preparations for custom font selection are archived as branch `xxx-characters`

PR #342, opened 2020-12-27
prevent each page from having its own "JobsPanel"

erwindon:
Most pages consist of 2 panels, with the right-hand panel a (short) list of recent jobs.
However, it is is technically a different panel each time.

There should only be one panel that is re-shown each time.

none

**Additional context**
This is only an issue since the migration to a Single Page Application (see #339).

Needs update of routing logic as this panel might appear on both the previous page and the next page.
The visibility of the panel is managed by `Router.js`.

### DIFF
--- a/saltgui/index.html
+++ b/saltgui/index.html
@@ -40,9 +40,6 @@
       <div id="warning" style="display: none"></div>
     </header>
 
-    <div id='route-container'>
-    </div>
-
     <div class='popup' id='popup-run-command'>
       <div class='run-command'>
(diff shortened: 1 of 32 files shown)


erwindon:
this is quite complicated due to the fact that the router manages PAGES, not PANELS.
serious GUI related problems will occur when panels are transferred from one page to another.
panels cannot be shared between pages because a panel includes its own DIV.
and since a DIV is an ELEMENT, it can have only one parent...

PR #326, opened 2020-12-29
fix job/state colors

erwindon:
**Describe the bug**
job/state colours are hardcoded and sometimes wrong.
different styles of green/yellow/red are used, based on the background.
but sometimes the wrong colour is used.

additionally, the colours are hardcoded instead of used indirectly through styles

**Expected behaviour**
use darker colours on white, use bright colours on black


erwindon:
fixed most of this with css-styles; cherry-picked from dev-branch (so no branch-merge)

PR #367, opened 2021-08-04
pls support eauth type sharedsecret

hoaivan:


erwindon:
@hoaivan 
technically, the PR is simple enough.

however, this method is not secure enough to just add to the baseline of SaltGUI.
I'll add it only based on a configuration parameter.
unfortunately, the `master` file is not available before login (otherwise we could have used the `external_auth` config to decide which login options to show in the first place).
the next alternative is a small file `saltgui/static/saltauth.txt` to be placed on the server.
the file will contain the list of additional auth modules that can be selected, to keep it generic.

the drop-down for authentication methods is divided in 3 parts. part 1 is for the recommended option and contains only `pam`. part 2 is for the additional options and currently holds `file`, `ldap`, `mysql` and `yubico`. part 3 is for the experimental methods and is only visible in git branches where a new method is introduced. your PR added `sharedsecret` in part 1.

the source code also contains a comment that mentions "sharedsecret". that comment must be updated too when this change would be applied.

don't bother to update this PR any further.
I'll make the changes and then ultimately just close this PR.

erwindon:
that first version was not a lot of work. this is what it looks like:

afbeelding

but some work is left:
* the file `salt-auth.txt` is loaded asynchronously and does not complete before the dropdown-box is created. the settings are thus applied the next time. need to work on refreshing the dropdown-box when new info comes in.
* since the 3 section (standard/other/salt-auth.txt) are visually similar, a lot of code duplication can be removed by reorganizing a bit.
* updating the documentation.

erwindon:
this PR is replaced by PR #372

PR #372, opened 2021-08-06
add support for additional salt.auth types

erwindon:
In #367, @hoaivan requests the addition of auth method `sharedsecred`.
However, that is not a safe mechanism unless additional measures are taken to prevent leaking these credentials.
Therefore `sharedsecret` will not be added as a standard option in the login screen.

But here is an alternative...
A file `salt-auth.txt` can be placed on the server, which contains all of the additional auth methods that should be shown.
So additional options will only be visible when they are added to that file.
And also, any other auth mechanism can be made selectable.

Note that an auth method does not work unless it is also configured in the `master` file.

This PR replaces the original PR #367.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ yarn.lock
 package-lock.json
 saltgui/static/minions.txt
 saltgui/static/mkMinions.sh
+saltgui/static/salt-auth.txt
--- a/README.md
+++ b/README.md
@@ -70,6 +70,12 @@ SaltGUI supports the following authentication methods supported by salt:
 - mysql
 - yubico
 
+Since pam by itself is already very powerfull, that one is mentionned as standard.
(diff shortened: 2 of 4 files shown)


erwindon:
@hoaivan can you please give the code from this PR a try?

erwindon:
@hoaivan I did not receive a reaction. can you still please try this PR?

hoaivan:
I will schedule my time for this.

erwindon:
@hoaivan it seems you are unavailable.
I've tested the updated login screen and:
* it works properly with `salt-auth.txt` absent
* it works properly with `salt-auth.txt` as an empty file
* it works properly with `salt-auth.txt` containing only `sharedsecret`
* it works properly with `salt-auth.txt` containing `sharedsecret` and `xyz` (though choosing `xyz` leads to nothing)

PR #377, opened 2021-08-21
enhance visibility and layout of buttons

erwindon:
buttons have poor layout: the spacing is ad-hoc
buttons have poor visibility: the button image is usually a visual character so that the active area is unclear

This PR creates a grey background colour for each button.
This immediately makes the layout problem more visible, so the layout instructions are re-worked.

### DIFF
--- a/saltgui/index.html
+++ b/saltgui/index.html
@@ -46,14 +46,14 @@
     <div class='popup' id='popup-run-command'>
       <div class='run-command'>
         <!-- 2716 = HEAVY MULTIPLICATION X -->
-        <div class='nearly-visible-button' id='cmd-close-button'>&#x2716;</div>
+        <div id='cmd-close-button' class='small-button small-button-right small-button-for-click'>&#x2716;</div>
         <!-- 2753 = BLACK QUESTION MARK ORNAMENT -->
         <!-- FE0E = VARIATION SELECTOR-15 (render as text) -->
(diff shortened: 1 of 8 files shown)


erwindon:
@dawidmalina 
In late 2018 and early 2019, you provided an updated design for SaltGUI.
This design has been still remains, but there were a few minor issues with the (menu) buttons.

Because, over time, a few more buttons were added to several screens:
* button for panel-level menu (until then: only a menu item existed per table row)
* button to start search in tables
* button to pause/continue automatic screen updates
* and a few more

And the buttons were not always recognizable since they were typically constructed using a single graphical character. So it was not always clear that the character even represented a button or a popup-menu. Also the extra margins around these characters were not visible. This resulted in an odd look for buttons that were near the edge of a block.

So a bit improvement was needed...

This PR cleans up the graphical design wrt to regular buttons and popup-menu buttons.

Can you please have a look at the graphical side of the result?
If/when you feel like it of course, as you may have moved away from SaltGUI or even away from SaltStack.

thx!

dawidmalina:
Hi @erwindon I can take a look on this but not within next 2 weeks. Never the less I must agreed with you - my involment with project, saltstack is close to zero right now as I am focusing now on different topics. I know that SaltGUI is in good hands :)

erwindon:
@dawidmalina thx for the quick response and the compliment! I'll now not let this PR depend on this, but feel free to still take a look at it at any time and send any remarks this way.

Issue #381, opened 2021-09-06
Installation Instructions


achimmihca:
Installation Instructions

I had some trouble installing SaltGui.

I would had wished for the following information in your README.md
- The [saltgui](https://github.com/erwindon/SaltGUI/tree/master/saltgui) subfolder contains the webapp and contains all necessary files
- No DB setup or similar is required, because SaltGUI always fetches the data fresh from the salt-master via a REST API
- salt-api (command line tool for SaltStack) and CherryPy (HTTP server) are used to serve the REST API and SaltGUI itself
    - to install salt-api run `sudo apt install salt-api` (this is already in the README.md)
    - CherryPy does not need additional installation
    - The location of the saltgui folder on your system is not relevant. However, the path must be set in rest_cherrypy config to be serverd correctly.
        - The recommended path is `/srv/saltgui`
- Using the following config, SaltGUI will be served by CherryPy when accessing http://localhost:3333 in your browser:
```
rest_cherrypy:
  port: 3333
  host: localhost
  disable_ssl: true
  app: /srv/saltgui/index.html
  static: /srv/saltgui/static
  static_path: /static
```
        
Note that README.md already contains the rest_cherrypy config. But it was not clear to me that SaltGUI itself is served by rest_cherrypy. I tried to place SaltGUI files in `/var/www` to serve them using Apache, which did not work.

Would be nice if you can include it in the docs.

Thanks!

erwindon:
@achimmihca thx! I will add most (or all) of this to the README.md. (that will be after next weekend due a busy schedule)

achimmihca:
Thanks!
I just noticed that I would change:

> Restart everything with `pkill salt-master && pkill salt-api && salt-master -d && salt-api -d`

to

> Restart everything with `pkill salt-master; pkill salt-api; salt-master -d; salt-api -d`

The && operator executes without waiting, and I don't think that there is a guarantee that pkill is executed before the start. So you might end up killing the new process.
Using ; guarantees that the commands are executed one after the other.

erwindon:
@achimmihca your latest remark is not valid...
there is a huge difference between `&` and `&&`
`&` starts the command in the background
`;` starts the command in the foreground
`&&` starts the command in the foreground, just like `;`, but only when the previous command succeeded

erwindon:
merged after review+correction

PR #382, opened 2021-09-12
improved documentation

erwindon:
See #381

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ external_auth:
 ```
 - See `docs/PERMISSIONS.md` for more restricted security configurations.
 - `saltuser` is a unix (PAM) user, make sure it exists or create a new one.
-- At the bottom of this file, also setup the rest_cherrypi server:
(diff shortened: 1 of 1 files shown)


erwindon:
@achimmihca
This is my PR with the implementation for #381.
I'll not recommend /srv/ as the actual top-level installation directory. But I added the suggestion to make /src/saltgui a soft-link to the actual installation directory.
Can you please indicate whether you think anything is missing, unclear or wrong?

achimmihca:
Thanks looks good.
I would add to the README the URL where saltGUI can be accessed when using the given config. Thus, I would replace

`At the bottom of this file, also setup the rest_cherrypi server:`

With

`At the bottom of this file, also setup the rest_cherrypi server to access SaltGUI from "http://localhost:3333":`

erwindon:
ok thx! I added a few extra words because the application is accessible on more than "localhost" alone with that configuration.

PR #429, opened 2022-03-12
Allow template definitions to be read from a file

erwindon:
Currently, templates can be defined in the `master` file.
But that has a disadvantage in 2 cases:
* changing a profile needs a salt-master and/or salt-api restart
* salt-api may be configured to run on a different host; in that case having setting in the master file should be avoided; this is because the purpose of separating salt-master and salt-api to to get a higher availability of the salt-master. making changes often and requiring a restart is opposite of that.

Add a function to read templates from a separate file `saltgui/static/salt-templates.json`.


erwindon:
implementation complete, started testing

PR #432, opened 2022-03-16
custom command help

achimmihca:
Button to show "custom command help", which is defined in salt master configuration

I very much like the command help from the '?' button.
Similarly, I want to be able to provide my own command help, with commands that we use frequently.
I know there are command templates. However, I wanted to provide a more general help.

- There is now a new button to show custom help text in the run-command-overlay
- Clicking the button fetches the salt config and creates the help text from config element `saltgui_custom_command_help`
    - Note that it is also possible to provide other help here, not necessarily related to commands.

For example, the following config:
```
saltgui_custom_command_help: |
  <h2>Job Commands</h2>
    runners.jobs.active
      => Show active jobs
      
    runners.jobs.list_job «JID»
      => Show job with given id
```

Results in the following help:
2022-03-16 11_10_55-SaltGUI

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ The version tagged `release` is the latest released version. The version `master
 - View the live events on the salt-event bus
 - View internal documentation for any salt command
 - View external documentation for any salt command
+- Define your own custom documentation for commands
 - Match list of minions against reference list
 - Match status of minions against reference list
 
@@ -190,6 +191,21 @@ saltgui_public_pillars:
     - pub_.*
 ```
 
+## Custom command documentation
+A custom HTML help text can be shown from the "Manual Run" overlay.
+
+Therefor,
+- specify saltgui_custom_command_help in the salt master config. Example:
+```
+saltgui_custom_command_help: |
+  <h2>Job Commands</h2>
+    runners.jobs.active
+      => Show active jobs
+      
+    runners.jobs.list_job «JID»
+      => Show job with given job id (JID)
+```
+- Hover the documentation icon (`📖︎`) near the command input field and select `Show custom help`
 
 ## Message-of-the-day
 A message-of-the-day (motd) can be added to the login screen.
--- a/README.md
+++ b/README.md
@@ -195,7 +195,7 @@ saltgui_public_pillars:
 A custom HTML help text can be shown from the "Manual Run" overlay.
 
 Therefor,
-- specify saltgui_custom_command_help in the salt master config. Example:
+- specify `saltgui_custom_command_help` in the salt master config. Example:
 ```
 saltgui_custom_command_help: |
   <h2>Job Commands</h2>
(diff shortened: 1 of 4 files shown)


erwindon:
I'll look into this PR next sunday...

erwindon:
The request is approved as a concept/idea.

I'm proposing the following instead (A):
* The `λ` button is replaced by a new menu-item in the documentation-drop-down-menu, which is already shown as 📖︎;
* The menu-item title is "Show site-specific help", or something better;
* The menu-item is only visible when there is site-specific-help available;
* The documentation will be extended to mention this feature.

Or (B):

* The `λ` button is removed;
* The existing '?' button shows the generic help-text (at the top) as well as the site-specific-help-text (at the bottom)
* The documentation will be extended to mention this feature.

Or (C):

* Same as (B), but additionally:
* Triangle icons are added to both parts of the help text to be able to collapse them (only strictly useful to collapse the existing help-text). The state of these icons remains constant during the session so that re-viewing the help-text shows the same layout as before.

erwindon:
@achimmihca 
any preference of comments?

achimmihca:
I prefer (A).

I want to keep the '?' help separated, because
- it shows context sensitive help
- the (from my perspective) more relevant custom help text would be easy to miss when mixed with the normal help text

achimmihca:
I implemented your suggestion (A).

However, I did not like "Show site-specific help".
I prefer "Show custom command help" because it is available from the "run command" overlay and site-specific would be weird because the end user has only one SaltGUI.

2022-03-22 09_24_24-SaltGUI

README.md
```diff
@@ -190,6 +191,21 @@ saltgui_public_pillars:
     - pub_.*
 ```
 
+## Custom command documentation
+A custom HTML help text can be shown from the "Manual Run" overlay.
+
+Therefor,
+- specify saltgui_custom_command_help in the salt master config. Example:
```
erwindon, line 198: use back-quotes around names of settings

erwindon:
review notes:
* the content of `saltgui_custom_command_help` is potentially too big to keep it in the session as organized by `_handleLoginWheelConfigValues`; so a separate use of `this.api.getWheelConfigValues` is acceptable
* the command-box is an odd construction, therefore some ugly constructions are acceptable

achimmihca:
I understand your review notes as "there is no reason why this should not be merged". Is this correct?

> normally, pull-requests are created from a branch. this one is from master. but it does not lead to problems.

It is from master branch **of a fork**. From my point of view, this works like a normal branch in the same repo.

erwindon (approved):

erwindon (approved):

erwindon:
tested OK

erwindon:
I've reverted this PR
there were this-vs-static changes that were not complete.
this cause the commandbox to fail in several cases (e.g. when initiated from a menu item)

erwindon:
I'll rebuild this change...

Issue #431, opened 2022-03-15
Redirect to original URL parameters after login


achimmihca:
**Is your feature request related to a problem? Please describe.**
First I want to say thank you!
SaltGUI is a great tool for us in our new configuration management.
Now, we are planning to add links to SaltGUI in notification E-Mails.

There is a little issue though:

- When clicking a link to SaltGUI (e.g. from E-Mail client), then SaltGUI opens with the login screen. This is ok.
    - Example: http://my.salt.com:3333/?id=20220315102657264588#job
- The problem is that after successful login, SaltGUI does not redirect to the clicked link.
    -  Instead, the URL after login is always http://my.salt.com:3333/#minions

**Describe the solution you'd like**
After successful login, SaltGUI should navigate to the URL that was used to open SaltGUI.
Maybe an implementation could look at the original URL parameters (query and fragment) and reapply them after successful login?

**Describe alternatives you've considered**
Current workaround:
- Log in to SaltGUI, then paste the URL of the link in the address bar of the browser, and hit enter.
    - This works because SaltGUI does not open the login screen again for an already logged in browser tab.

erwindon:
approved

erwindon:
The typical implementation for this is that the original url and parameters become parameters of the login screen during the redirect to the login screen.
e.g.:
http://localhost:8000/?id=20220312142559354840#job
(when not logged in) redirects to something like:
http://localhost:8000/?redirect=job&id=20220312142559354840&reason=no-session#login
the parameters are ignored by the actual login procedure (except for `reason`), but they are used to construct the proper target url to go to after finishing the login procedure.

note that there is always the risk that the original page that was specified:
* does not exist
* exists, but has insufficient parameters
* exists, but has wrong parameters
* exists, but does not appear in the list `saltgui_pages` (which fact should remain ignored)

achimmihca:
Sounds good.

However, I have an issue with salt itself: the JID does not seem to be available when runnning a script.
Thus, I cannot create links to SaltGUI.

Anyway, the redirect feature in SaltGUI is still useful when sending/embedding links manually.

erwindon:
See #434 
Can you please review/test the PR?

erwindon:
merged

erwindon:
unmerged

PR #434, opened 2022-03-23
Redirect to original URL parameters after login

erwindon:
See issue #431

### DIFF
--- a/saltgui/static/scripts/Router.js
+++ b/saltgui/static/scripts/Router.js
@@ -260,8 +260,11 @@ export class Router {
       // the fact that we don't have a session will be caught later
       // but this was shows less error messages on the console
       // but do not destroy the reason when login is already the goal
+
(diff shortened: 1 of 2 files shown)


erwindon:
@achimmihca
This is the PR for #431
Can you please test this PR?
(at this moment, I cannot assign this one to you, that should be possible after you leave at least one comment here)

achimmihca:
Works great!
Minor issue (only nagging): `reason=no-session` is still in the URI after successful login.
Maybe one could remove this on redirect after successful login.

erwindon:
Thx for testing!

> Minor issue (only nagging): reason=no-session is still in the URI after successful login.

That's not nagging. That is valuable feedback. Fixed that.

Can you test once more: both the original intention and the additional fix?

achimmihca:
Retestet, works like a charm now.
Many thanks!

PR #436, opened 2022-03-25
custom command help - part deux

erwindon:
See also #432, which had technical problems
this is an exact re-implementation

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,7 +205,7 @@ saltgui_custom_command_help: |
   <h2>Job Commands</h2>
     runners.jobs.active
       => Show active jobs
-      
+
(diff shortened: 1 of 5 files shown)


erwindon:
decided to treat the custom-help-text as any other global setting and read it just after login.
(actually it was already collected but just not stored)
from then on the custom-help-text is available as a property in the session.
that makes it very easy to use...

erwindon:
@achimmihca FYI
this PR replaces your PR (and it also replaces the issue that I created unnecessarily).
can you please confirm that this implementation of "custom-help" works for you?

achimmihca:
Yes, works as desired. Thanks!

PR #214, opened 2022-03-26
should always use full_return

erwindon:
Some commands return only very simple information, like true/false. This is indistinguishable from the return of `false` on a timeout.

The API should use `full_return` as much as possible. The response from that clearly indicates success/failure as well as a separate result.

Update: while the data is better separated from the meta-data, it does not indicate success/failure. instead, it provided a result-code separate from the results. but the indication for success/failure is still missing.

Update: this is only relevant for free-format commands, not for the internal commands. When internal commands fail we don't show much extra information anyway.

Update: this must be arranged as a setting `saltgui_use_full_return: True` (default: `False`) as this is a breaking change. not only will the gui show different output, also the job history as maintained by the SaltStack system will change a little bit. And that may be an unpleasant surprise for some.


erwindon:
retrieving setting is done
TODO: use setting in many places
TODO: testing

erwindon:
with full_return, also the JID is returned.
this allows SaltGUI to jump directly to the JOB view for that

PR #430, opened 2022-03-31
SaltGUI shows wrong start time in job list

erwindon:
**Describe the bug**
The "Start Time" column of Jobs in SaltGUI is off by one hour.
This is probably related to the unsed time zone.

The time is shown correctly in the Salt output. Thus, I think the problem is with SaltGUI.

**To Reproduce**
- We have a time zone set on our server: `sudo timedatectl set-timezone Europe/Paris `
- When running a job, then salt shows the correct start time, for example:
```
done, 1 response
my-minion:▽●

✔echo "Hello Bob" id=say_hello (from hello.sls)
    Function is cmd.run
    Command "echo "Hello Bob"" run
    pid: 14384
    retcode: 0
    stderr: ""
    stdout: "Hello Bob"
    Started at 10:36:48.076396
    Duration 1.172 s
1 succeeded, 4 changes
```

- When viewing this Job in SaltGUI, the "Start Time" column says: `2022, Mar 14 09:36:44.728128`, which is off by approx. one hour

**Expected behaviour**
 The "Start Time" column in SaltGUI should show the same value as the output from salt.

Is there maybe a time zone setting in SaltGUI to fix this?


erwindon:
(the issue is converted to a PR from my repo)

* datetime (or time) fields are now re-formatted in the local representation, but still utc
* the representation includes the preferred decimal separator (see https://en.wikipedia.org/wiki/Decimal_separator)
* fields show additional text to show the local time with the abbreviated timezone, e.g. "31-3-2022 13:37:27,454226 (15:37:27 CEST)". The date may have also be different, but that is not shown.

@achimmihca can you take a look at this intermediate result?

TODO:
add a setting in the master file on how to format the date/time fields
add a section in the Options page to show that setting
most likely, the variations are:
* utc (no TZ visible) this is the old default, remains the default
* local (no TZ visible)
* utc with local-time (TZ visible)
add a tooltip with representation for both local and utc, also shows the full amount of fractional digits for seconds
the default will be the OLD representation.
the development branch currently has one of the new representations.

erwindon:
still working on lower priority stuff in the same area:
* use the new datetime logic in the "Options" screen
  * options-panel: for session_start and session_expire
  * stats-panel: for "Current Time", "Start Time" and "Uptime" on top-level
  * stats-panel: for "Current Time", "Start Time" in requests
  * stats-panel: for "Current Time", "Start Time" in slow-queries
* use the new datetime logic also in highstate overviews (start-time)

I cannot do much work next week, so progress on the extra items will be slow...

achimmihca:
I just tested this. I get a tooltip that shows the UTC and local time.
But I get the same result no matter how I set `saltgui_datetime_representation`. The column always shows utc and the tooltip the utc and local values.

What I expected is: The column shows the local time and the tooltip optionally the utc time as well.

But still: local time in tooltip is correct.
I guess the rest will be implemented when you find time?

Thanks!

2022-04-04 09_52_10-SaltGUI

erwindon:
@achimmihca 
the content of the tooltip is always the same (utc+local, with full time resolution), so that is correct here.
all datetime representations are with the local rules (e.g. dot between day and month and comma as decimal separator, both as usual in your country), so that is correct too.
You should be able to choose the application datetime-format using the master configuration file or overrule using the Options page. I will look into that part in a few days...
Thx so far!

erwindon:
@achimmihca 
I fixed the setting of the date+time representation.
You should be able to select between the 4 variations again.
I simplified the values of the setting, replacing the previous `+` with a `-`, so that the values can be used also as part of an html-element-id.

Can you please try again?

erwindon:
@achimmihca 
just for completeness, can you try again with unmodified code?

achimmihca:
@erwindon , I don't see any change?
`JSON.stringify` is still there.

Did you push anything?

erwindon:
> Did you push anything?

ehh, no, somehow missed that, done that now...

PR #439, opened 2022-04-08
removed minions still show up in issues tab

erwindon:
**Describe the bug**
Removed minions are still in the issues panel

**To Reproduce**
1. add a minion.
2. accept its key
3. run highstate. Make it fail. (other failed state may work as well. haven't tested that)
4. remove minion
5. open issues tab. Minion still mentioned there as failed.

**Expected behavior**
The minion is gone. I expect it to be gone everywhere :)

**Additional context**
If it's just a matter of time before it's erased then there's not really an issue so just close it then.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,27 @@ saltgui_templates:
         command: test.version
 ```
 
+When there are a lot of templates, they can be organized into categories.
+e.g.:
+```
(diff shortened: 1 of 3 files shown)


erwindon:
I goofed!
This PR is pointing to the wrong branch...

PR #438, opened 2022-04-08
(sub)folders for job templates

erwindon:
My templates list is getting a bit long

I'd love to be able to sort the job templates a bit better instead of just one big list. One way would be in collapsible folders in the same way as you can open and close the job results per minion.

I don't really know of another way. If you can think of another way to order or categorize them better, also great..

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,27 @@ saltgui_templates:
         command: test.version
 ```
 
+When there are a lot of templates, they can be organized into categories.
+e.g.:
+```
+saltgui_templates:
+    template1:
+        description: First template
+        target: "*"
+        command: test.fib num=10
+        category: cat1
+    template2:
+        description: Second template
+        targettype: glob
+        target: dev*
+        command: test.version
+        categories:
+          - cat1
+          - cat2
+```
+When at least one template is assigned to a category, then you can select a template category before
+selecting the actual category. Otherwise that choice remains hidden. Templates can be in multiple categories
+when a list of categories is assigned.
 
 ## Jobs
 SaltGUI shows a maximum of 7 jobs in on the right-hand-side of the screen.
--- a/saltgui/static/scripts/panels/Panel.js
+++ b/saltgui/static/scripts/panels/Panel.js
@@ -154,14 +154,36 @@ export class Panel {
     table.id = this.key + "-table";
     table.classList.add(this.key);
 
+    let anyHiddenColumns = false;
+    if (pColumnNames) {
+      for (const colName of pColumnNames) {
+        if (colName.startsWith("@")) {
+          anyHiddenColumns = true;
+        }
+      }
+    }
+
+    if (anyHiddenColumns) {
+      for (const colName of pColumnNames) {
+        const col = document.createElement("col");
+        if (colName.startsWith("@")) {
+          col.style.visibility = "collapse";
+        }
+        table.append(col);
+      }
+    }
+
     if (pColumnNames) {
       const thead = document.createElement("thead");
       thead.id = this.key + "-table-thead";
       const tr = document.createElement("tr");
       tr.id = this.key + "-table-thead-tr";
 
-      for (const columnName of pColumnNames) {
+      for (let columnName of pColumnNames) {
         const th = document.createElement("th");
+        if (columnName && columnName.startsWith("@")) {
+          columnName = columnName.substr(1);
+        }
         if (columnName && !columnName.startsWith("-")) {
           th.innerText = columnName;
         }
--- a/saltgui/static/scripts/panels/Templates.js
+++ b/saltgui/static/scripts/panels/Templates.js
@@ -12,7 +12,7 @@ export class TemplatesPanel extends Panel {
 
     this.addTitle("Templates");
     this.addSearchButton();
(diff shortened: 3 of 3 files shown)


erwindon:
converted into a PR

erwindon:
> the current list is built with html-input-list, which does not support collapsible hierarchy

That was a false statement: the html-input-list is used for node-groups, not for templates.

erwindon:
DONE:
* the Templates panel now shows an overview including the assigned categories
* the "Category" column is hidden unless there are actually templates that have a category
* the "Category" column is renamed to "Categories" when there are templates that have multiple categories

TODO:
* Show a dropdown menu for the template-categories. This is a bit ugly, as the category selection and the template selection will be in 2 separate drop-down menus. The menu-system in SaltGUI does not support sub-menus. The menu will be hidden when no categories are assigned.
* Add artificial template category "(none)" for templates that have no category
* Add artificial template category "(all)" for all templates, this is the default.
* Reduce the templates dropdown menu to only the items of the selected category.

erwindon:
@xzenor
A proposed implementation for this request is ready in this PR.
--> Can you please give it a try? Please read the updated manual for details.

There is an alternative visualization in my mind: use a single drop-down menu, which first has the categories as menu-items, followed by the menu-items for the templates that are matching the chosen category.
nice when there are only a few categories, but otherwise maybe a bad idea?

A second alternative visualization: use nested menus. this is a bit of work since SaltGUI currently does not support nested menus. in this alternative, the category cannot be selected. instead, each category shows the list of matching templates in a sub-menu.

small improvement will follow soon: the selected category will be made visible outside the menu.

xzenor:
I'll give it a try today

xzenor:
Assigning the categories works great! The visualization isn't really solving the problem though. It has an extra column now that I can sort on but that's it, right? Or am I looking at cached content?
 
3rd idea, a list of toggles at the top to enable or disable categories? Might get a little messy when you have a lot of them though so maybe that's not the best idea either...
I Like your first 'alternative' visualization, actually..

erwindon:
> It has an extra column now that I can sort on but that's it, right?

The "Templates" page only shows the available templates with all their details. The only change there is the addition of the category column (but only when at least one template has a category assigned).
On this page, the traditional sorting (click column title) and filtering (click looking glass) are available as before.
No further changes were made there.

The second part of the change is in how the templates are used.
I did not realize that templates can also be used by just clicking on them on the Templates page.
The templates are available for use in the commandbox popup. A drop-down menu just after its title "Manual Run" holds all the references to the templates. On that location, a second menu was added (when needed) to select the category first.

erwindon:
> 3rd idea, a list of toggles at the top to enable or disable categories? Might get a little messy when you have a lot of them though so maybe that's not the best idea either...

so I'll ignore it...

> I Like your first 'alternative' visualization, actually..
this also has a problem when there are many categories.

I'll keep the current implementation for now.
Any improvement on this can be done later.

xzenor:
> so I'll ignore it...  

I just mentioned it because maybe it would spark an improvement on that idea with you.   

I can now sort by category. that's nice.

xzenor:
> The second part of the change is in how the templates are used.
I did not realize that templates can also be used by just clicking on them on the Templates page.
The templates are available for use in the commandbox popup. A drop-down menu just after its title "Manual Run" holds all the references to the templates. On that location, a second menu was added (when needed) to select the category first.  

Completely missed this! This is awesome!

erwindon:
please let me know whether anything needs to change for:
* the templates page
* the command box
* the documentation
* or anything else for this issue/pr...

xzenor:
A little pulldown to just show a certain category on the templates page would be nice.. Just like on the Manual Run page..

erwindon:
> A little pulldown to just show a certain category on the templates page would be nice.. Just like on the Manual Run page..

That is something that would be useful on other pages as well. e.g. filter minions by their OS or os-version, filter keys by their state, etc., etc.. So this would have to be made as as generic facility. This also means that a separate github-issue is best.

But..., the text-search facility that is available on all panels and it is very generic. So, the benefit of adding a category-search would be small.

A little more help may still be offered to the user though. See e.g. the Keys page. On that page, the search box provides a bit of help by suggesting the possible values for a key's state (accepted/unaccepted/denied/rejected). The same could be done on the Templates panel by providing the category names in the LoV for the search field. The search would not be limited to the category column only, but usually that would be barely noticeable. (unless you use category names like "a" or "b"!).

erwindon:
the search box on the templates panel now has the list of category names in its dropdown menu (this menu is part of the generic LoV facility for text fields).

xzenor:
Oh wow. I always thought that was just browser form-caching but you actually fill those. That definitely works!

erwindon:
for completeness:
* target field in command-box has list of node-groups as LoV, plus list of minions, plus 1 extra: "##connected"
* search field in keys screen has the 4 possible key status values as LoV
* search field in jobs screen has 6 text fragments that I thought are useful
* search field in job screen has 4 text fragments that I thought are useful
* (new) search field in templates screen has list of real categories as LoV

xzenor:
I've got nothing more to complain :)

erwindon:
I'll merge this one later today

xzenor:
Thanks for all your work man.

PR #454, opened 2022-06-09
Create codeql-analysis.yml

bsikar:
You should create code analysis.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,72 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


erwindon:
@bsikar 
before anything like this is added, please explain how it should be used

bsikar:
So code analysis will allow you to see under the security tab security vulnerabilities in your code. From github: [https://github.com/features/security](https://github.com/features/security). This will allow you to see any possible issues with your code.

For instance, when I added this to my fork, I saw this:
image

Now you are able to dismiss these issues, but it will at least bring them to your attention.

erwindon:
the finding from the screenshot alone is already worth something!
Since this is a github feature, I'll try it out first. Looks like I have to hit some buttons in section "Code security and analysis"...

bsikar:
Yes you will and this file is generated by github, so I can close this pr and you can do it on your end. The file will look exactly the same.

erwindon:
please keep the PR open as it is also a reminder of work at hand

the problem with larger configuration files like this, is that it is unclear whether each of the items in it are just the defaults or not.

bsikar:
Ah I understand, feel free to close it once you get it sorted out, I will keep it open.

erwindon:
fyi: note that the following tools are also regularly use by me:
* npm run stylelint
* npm run eslint

erwindon:
fyi: I just fixed the 2 security warnings:
* the first one was just a coding mistake (see also the above screenprint)
* the second one was a false positive, which I have now suppressed

PR #465, opened 2022-08-03
Enhanced Highstate page

erwindon:
I would like a enhanced hightstate page on which we could define some display parameters.

the enhanced page would allow (at least) selection of the salt-commands to consider for the page :
- with the choice "state.highstate+state.apply" it would have the current functionality
- with the choice "state.highstate" it would be limited to that information only
- with the choice "state.apply" it would be limited to that information only

**Additional context**
we only use "state.highstate" to define whether the server is in highstate or not, and "state.apply" is used only for deployment. So this new page could be very useful. 
see also : https://github.com/erwindon/SaltGUI/issues/458

### DIFF
--- a/saltgui/static/scripts/panels/HighState.js
+++ b/saltgui/static/scripts/panels/HighState.js
@@ -24,6 +24,9 @@ export class HighStatePanel extends Panel {
     this.addPanelMenu();
     this._addMenuItemStateApply(this.panelMenu, "*");
     this._addMenuItemStateApplyTest(this.panelMenu, "*");
+    this.addSettingsMenu();
+    this._addMenuItemUseStateHighstate();
+    this._addMenuItemUseStateApply();
     this.addSearchButton();
     this.addPlayPauseButton();
     this.addHelpButton([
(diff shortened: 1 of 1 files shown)


erwindon:
migrated _issue_ to _PR_

erwindon:
@pju51 
the "highstate" page now has the option to select for the use of each of state.apply, state.highstate.
the choice is remembered in the browser memory, but outside the session, so that it does not need to be applied again-and-again.
can you please take a look at this?
(I know that your availability in the next weeks is very low, take your time)

erwindon:
@pju51
welcome back!

> still in draft or I can test ?
It is draft because it is still being tested (by you), I'm not currently changing anything. this is mainly a precaution, so that I do not accidentally merge.

same information, different wording:
I updated the "highstate" page.
An option was added to select whether the results from state.apply and/or state.highstate are used to present the data.
This choice is not configurable in the /etc/salt/master file, but it is remembered in the browser that you use.

erwindon:
note that I just rebased again, use "git pull -r" to update

pju51:
Testing..

Highstate page only show the last result ?

Example:
I execute on a server a **state.apply** then a **state.highstate** for X or Y reasons. I can only see the result for **state.highstate** ?

from my side, each day a **state.highstate** (in test mode) is executed to check if we have some server not in highstate, Is it possible to keep this information ? it allows me to monitor my servers and check their status.

erwindon:
> Highstate page only show the last result ?
yes
for each minion, the latest state is shown, but based only on the selected command types, so `state.apply` and/or `state.highstate`, that you select in the panel-settings-menu (the one with the cog). starting with this branch, menu's that have only settings are marked with a cog, other menus still have the hamburger icon.

SaltGUI currently never makes a distinction between regular jobs and jobs with `test=true`. It just does not even look at that parameter.
Should the menu include choices to filter for `"test=true"` and `not("test=true")` on this page?

pju51:
ok so the filter works like a charm.  
not needed to a filter for "test=true" for me.

erwindon:
note: the information from https://pythontechworld.com/issue/saltstack/salt/62410 was helpful to fix my development machine

PR #458, opened 2022-08-03
ignore some jobs on highstate page

erwindon:
Hello,
it's not a bug, but questions or improvements.
I am using the separate SaltGUI (1.28.0) solution in a nginx container (nginx:1.21.6). 
Everything works fine but I have 2 problems with the hightstate page:

1) my saltstack master uses a Gitlab as a backend, so to deploy a specific branch with salt, I use this command on the host as an example: `salt-call state.apply saltenv=XXX pillarenv=XXX test=true`  (XXX = git branch)
The problem is that the result of the command will also be displayed on the saltgui hightstate page, but we should only see the results of the master branch and not the result of my XXX branch, because only the master branch is important to define if the server is in highstate or not.
Is it possible to hide these parasitic results in order to keep only the results of the master branch? or any other solution (like to show only the **state.highstate** and hide everything else) ?

2) I have a lot of minions and states, and in the highstate page,  the result of the highstate is far too long:
Is it possible to have only the final result ? and the detail if we click on the green/yellow/red circle ? or another proper solution.
I already tried this solution: https://github.com/erwindon/SaltGUI#performance,  but I didn't see any change, maybe because I use a separate SaltGUI.
**Screenshots**
image

thanks a lot.

NOTE: nr. 2 is now in #459 
NOTE: question about salt-syndic (below) is now in #460

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -237,6 +237,30 @@ saltgui_public_pillars:
     - pub_.*
 ```
 
+## Highstate
+The highstate page provides an overview of the minions and their latest state information.
+At most 10 highstate jobs (`state.apply` or `state.highstate`) are considered.
+
+Individual low-states can be re-tried by clicking on their state symbol.
+Note that since the output of the `state.sls_id` commands is not considered in this overview,
+the result of such action has no effect on this screen.
+Use `state.apply test=true` to update the information without making changes on the minions.
+
+For organisations where the 'saltenv' facility is used, it is possible to limit the jobs that are
+considered to include (or exclude) only specific saltenvs.
+e.g.:
+```
+saltgui_show_saltenvs:
+    - base
+```
+or
+```
+saltgui_hide_saltenvs:
+    - env2
+    - env3
+```
+Typically only one of these variables should be set.
+Jobs that were started without the `saltenv` parameter are always shown.
 
 ## Custom command documentation
 A custom HTML help text can be shown from the "Manual Run" overlay.
--- a/docs/README.md
+++ b/docs/README.md
@@ -260,7 +260,7 @@ saltgui_hide_saltenvs:
     - env3
 ```
 Typically only one of these variables should be set.
-Jobs that were started without the `saltenv` parameter are always shown.
+Jobs that were started without the `saltenv` parameter are, for this purpose only, assumed to use the value `default` for this parameter. This allows these jobs to be hidden/showed using the same mechanism. SaltGUI does not replicate the internal logic of the salt-master and/or the salt-minion to determine which saltenv would actually have been used for such jobs.
 
 ## Custom command documentation
 A custom HTML help text can be shown from the "Manual Run" overlay.
--- a/saltgui/static/scripts/CommandBox.js
+++ b/saltgui/static/scripts/CommandBox.js
@@ -438,8 +438,7 @@ export class CommandBox {
     while (targetList.firstChild) {
       targetList.removeChild(targetList.firstChild);
     }
-    const nodeGroupsText = Utils.getStorageItem("session", "nodegroups", "[]");
-    const nodeGroups = JSON.parse(nodeGroupsText);
+    const nodeGroups = Utils.getStorageItemObject("session", "nodegroups");
 
     const optionConnected = document.createElement("option");
     optionConnected.value = "##connected";
@@ -451,7 +450,7 @@ export class CommandBox {
       targetList.appendChild(option);
     }
 
-    const minions = JSON.parse(Utils.getStorageItem("session", "minions", "[]"));
+    const minions = Utils.getStorageItemList("session", "minions");
     for (const minionId of minions.sort()) {
       const option = document.createElement("option");
       option.value = minionId;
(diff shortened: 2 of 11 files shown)


erwindon:
migrated _issue_ to _PR_

pju51:
@erwindon, I restarted from 0, so I cloned your repo and I checkout to your latest **highstate-envs** branch,  but it still doesn't work on my side :

- test 1 with the key work "base":
image
image

-test 2 with the key work "master":
image
image

I don't see what additional testing I can do to help debug

PS : I'm going on holiday for 3 weeks, so I won't be able to test anything until then.

erwindon:
ok, that gives me 3 weeks to figure out what is going on... enjoy your holiday!

erwindon:
not related to the investigation:
I've updated the code to use the value "default" when no saltenv is given. This makes it possible to hide/show these jobs in the same way as jobs that do have the saltenv parameter. When you want to see these jobs and are using the `saltgui_show_envs` parameter, then you must now add the value "default" there (previously these jobs were always shown).

erwindon:
@pju51 
I did not find anything wrong with the 2 examples that you provided. Because the given saltenv is "CICD-1776" it is not matched by the given values "base" or "master" and so they are not shown here.
Since the images indicate that the jobs are shown for you, there must be something different.

The first reason might be that you are using old SaltGUI code. Not because you are in the wrong branch, or that the branch is not up-to-date, but because the browser has cached an old version of the code ==> use SHIFT-refresh.

I cannot think of other reasons at this time, so let's investigate.
I've added 2 lines of code that log information to the console of the webbrowser (git pull to get them)
1) the values of the parameters saltgui_show_saltenvs / saltgui_hide_saltenvs is shown. Note that the prefix "saltgui_" is already removed at this point.
2) for each job, the value of the saltenv that is used, or the value "default" when none was found in the job.
please let me know the results here

pju51:
Hello @erwindon ,

it was wrong, because I added :
**saltgui_show_saltenvs: base** (and master after for test)

so normally it should exclude from display all commands that do not use the branch base (or master) but in my case it displays them. I don't use an old code, I pull from your repo. Otherwise I didn't understand what **saltgui_show_saltenvs** is for.

from what I understood the parameter **saltgui_show_saltenvs:** allows to display only the commands which use the specified env?

maybe because of the browser cache, so I will try again with your new code + and cache refresh between each tests.

erwindon:
@pju51
welcome back!

In case you are wondering... I just forgot to mark this one as 'draft'  :-)

(and no rebase was needed here)

pju51:
I pulled the branch again, I tried in private mode on my browser (+ SHIFT-refresh to be sure)

My saltstack master is configured with "saltgui_show_saltenvs: master" and salt services (salt-api, salt-master) has been restarted.
image

I executed this on saltgui:
image

and the result :
image

I added a pillar env  but the result still appear on the highstate page when it should not.
what should I see in the browser console ??

pju51:
humm very weird...
currently it doesn't show me any commands in the hightstate page, as if it was filtered and I haven't changed anything from my last test.
Also, I always see the cmds on the right panel (recents jobs) in the high status page.

I will test with this parameter 
saltgui_show_saltenvs: base

erwindon:
@pju51 
let's clear one confusion...
the filter works on the parameter `saltenv=` that was included in the command.
therefore, the only reasonable setting is: `saltgui_show_saltenvs: SOP-4122`, with exactly the same value as it was used in the command.

with the code in this branch, the browser console should show 2 types of lines:
1) one line with an overview of the parameters `saltgui_show_saltenvs` and `saltgui_hide_saltenvs` as they will be used by SaltGUI. both should be array values. when a single string was configured as value, it will be shown as a 1-element array here.
2) one line with the JOB-ID and the SALTENV per job that was found (each job may apply to multiple minions).

Using `saltgui_show_saltenvs: base` will only find the jobs that really have a parameter `saltenv=base`. SaltGUI cannot know what the default saltenv is, even though in 99% of the cases it would be `base`, it can even be different per minion.
Using `saltgui_show_saltenvs: master` will most likely not be useful since that is a git branch-name, which neither SaltStack, not SaltGUI has any knowledge of.
Using ``: default` will catch the jobs that do not have a parameter `saltenv=` (or a parameter `saltenv=default`). That was an addition I made just before your holiday.

can you please specify a single combination that is NOT working?

pju51:
it's simple.

if I add this :
`saltgui_show_saltenvs: master`
I would like to see only the result of this command in the highstate page :
` salt-call state.apply test=true` (with or without  test=true)
 
and  if I try with a saltenv parameter like : 
` salt-call state.apply saltenv=XXX pillarenv=XXX test=true`  (with or without  test=true)
 it should not be displayed.

erwindon:
@pju51
> if I add this :
> saltgui_show_saltenvs: master
> I would like to see only the result of this command in the highstate page :
> salt-call state.apply test=true (with or without test=true)

> and if I try with a saltenv parameter like :
> salt-call state.apply saltenv=XXX pillarenv=XXX test=true (with or without test=true)
> it should not be displayed.

I cannot literally add the first part to SaltGUI as the word `master` is the name of a git branch.
But I already implemented the same functionality, using the keyword `default`.
Except for this detail, it should work like you expect.

pju51:
Thanks @erwindon  I will test soon.

I was a little busy previous weeks

pju51:
Hello @erwindon 

so I tested and....
it works like a charm with the keyword `default` ! 💯 

This MR is done for me, I validated and you can merge ! Thanks a lot

PR #459, opened 2022-08-03
Add the possibility to collapse the result of the states in the hightstate

erwindon:
Hello,

I am using the separate SaltGUI (1.28.0) solution in a nginx container (nginx:1.21.6).
Everything works fine but I have a problem with the hightstate page:

I have a lot of minions and states, and in the highstate page, the result of the highstate is far too long:
Is it possible to have only the final result with one circle ? and the detail if we click on the green/yellow/red circle ?
or another proper solution to reduce the result ?

I already tried this solution: https://github.com/erwindon/SaltGUI#performance, but I didn't see any change, maybe because I use a separate SaltGUI or maybe its not related with.
Screenshots

image

thanks a lot

### DIFF
--- a/saltgui/static/scripts/Character.js
+++ b/saltgui/static/scripts/Character.js
@@ -9,6 +9,7 @@ export class Character {
     // VS16 = request emoji representation
     Character._VARIATION_SELECTOR_16 = "\uFE0F";
 
+    Character.MULTIPLICATION_SIGN = "\u00D7";
     Character.NO_BREAK_SPACE = "\u00A0";
     Character.NON_BREAKING_HYPHEN = "\u2011";
     Character.HORIZONTAL_ELLIPSIS = "\u2026";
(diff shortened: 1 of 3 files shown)


erwindon:
migrated _issue_ to _PR_

erwindon:
@pju51 
I've added the required code to this branch.
Can you please take a look at the result?
Just go to page "highstate" and see.

pju51:
Hello @erwindon 

I confirm its okey. Thanks a lot

image

image

erwindon:
thx! 
I'll merge this soon.

erwindon:
final adjustments:
* made the max-length a static configuration parameter, still maximum is 20
* added 1 line in the page tooltip about switching to summary after 20 tasks
* made the background for the task-icon black, same reason as before: the visibility of the "task-changed" icon, which is yellow, which would be yellow-on-white. this also needed adjustment of the margins for the task-icons as there are now more visible.

PR #460, opened 2022-08-03
SaltGUI and saltstack multimaster infrastructure

erwindon:
Hi,
currently SaltGUI connects to a single master to execute salt commands, retrieve information (grains, pillars,..) and results.

In my case, we have 2 salt masters servers and 1 salt manager server, the manager manages the masters via salt-syndic installed on the 2 salt masters, and thus load balances the deployment requests to one or the other master.

the result is that if the manager sends the deployment to the server not "linked" to SaltGUI, I don't get the result, so basically I only get 50% of the saltstack results.

Is it possible to make SaltGUI compatible with a multimaster infrastructure?


erwindon:
migrated _issue_ to _PR_

PR #464, opened 2022-08-03
syndic: add warning to Keys screen when syndic is in use

erwindon:
When salt-syndic is in use, there is a difference between the list of keys and the list of minions.
Since the list of keys is then shorter, the user may have the feeling that keys are missing.

When salt-syndic is in use, show a warning about the possible limited list of keys.

**Additional context**
solution is to read configuration from master file
(previous solution mentioned reacting on specific salt-events)

### DIFF
--- a/saltgui/static/scripts/Api.js
+++ b/saltgui/static/scripts/Api.js
@@ -433,6 +433,8 @@ export class API {
       } else if (tag.startsWith("salt/job/") && tag.includes("/prog/")) {
         // progress value (exists only for states)
         CommandBox.handleSaltJobProgEvent(tag, data);
+      } else if (tag.startsWith("syndic/")) {
+        pRouter.keysPage.handleSyndicEvent();
       }
 
       pRouter.eventsPage.handleAnyEvent(tag, data);
--- a/saltgui/static/scripts/pages/Keys.js
+++ b/saltgui/static/scripts/pages/Keys.js
@@ -26,4 +26,8 @@ export class KeysPage extends Page {
   handleSaltJobRetEvent (pData) {
     this.jobs.handleSaltJobRetEvent(pData);
   }
+
+  handleSyndicEvent () {
+    this.keys.handleSyndicEvent();
+  }
 }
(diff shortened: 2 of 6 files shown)


erwindon:
migrated _issue_ to _PR_

pju51:
Hello @erwindon ,

still in draft or I can test ?

erwindon:
@pju51
welcome back!

> still in draft or I can test ?

It is draft because it is still being tested (by you), I'm not currently changing anything. this is mainly a precaution, so that I do not accidentally merge.

note that I just rebased again, use "git pull -r" to update

pju51:
@erwindon Hello,
where should I find the warning message? it doesn't appear anywhere

erwindon:
I was busy setting up many test-VMs to test all combinations. That was a lot of work and I only got halfway.
Auto-detecting the exact role of a salt-master within a salt-syndic network is too much work, it is unreliable and there is no uniform convention to describe the roles. I even doubt whether there is enough data available to do full auto-discovery.
I think I will simplify the implementation.
Instead, I'll just display the following indicators:
* for variable `syndic_master`
e.g. "The syndic-master of this salt-master is `hostname`." when the variable is set.
* for variable `order_masters`
e.g. "This salt-master is ready to work with salt-syndic nodes." when the variable is true.
* for the presence of syndic-events
e.g. "Events related to salt-syndic are seen in the salt-event-bus." when such events are seen.
* general warning
e..g. "Salt-syndic is in use in this network. Results may be received for minions that do not have accepted keys on this salt-master." when any of the indicators are present.

erwindon:
I've applied the simplification.
A warning will be visible on the KEYS page when the current salt-master is configured for use in a syndic topology.
I've added a new git-issue to show information on for minions that are in a multi-master setup, see #480.

Does SaltGUI show the proper information in your case? (with this git branch)

pju51:
I have the message : 
image
and its clear :)

so you can merge your branh. Thanks.

erwindon:
ok, thx!

PR #474, opened 2022-08-20
allow job output to be reformatted

erwindon:
formatting of job output is configured statically and can (for the current session only) be changed.
this makes it difficult to see other representations (including the original data()

Add option to select output type.

Also for the commandbox?
no, that is already a complex UI, further analysis should be done in the job-view

What about the other formatting options?
* doc yes/no
not relevant, this is only applied for output in the commandbox
* highstate saltgui/traditional/no
must be included
* state_output
"full", "terse", "mixed", "changes", "full_id", "terse_id", "mixed_id", "changes_id"
* state_output_pct
* state_verbose
* state_compress_ids, see also #478

### DIFF
--- a/saltgui/static/scripts/panels/Job.js
+++ b/saltgui/static/scripts/panels/Job.js
@@ -1,7 +1,9 @@
 /* global */
 
+import {Character} from "../Character.js";
 import {DropDownMenu} from "../DropDown.js";
 import {Output} from "../output/Output.js";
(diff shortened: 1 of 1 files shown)


erwindon:
TODO:
* fix sonar issue(s)
* make output selection only a temporary effect

erwindon:
implemented ok
but not suitable for the public yet
so, keeping it as branch `zzz_outputformat`

PR #479, opened 2022-09-16
Preserve API_URL in response

erwindon:
Hi, 

first of all I would like to thank you for your great GUI

I have SaltGUI with API_URL set as `/saltgui`. So I type  `http://example.com/saltgui` in my browser to log in. 
But after logging in I see `http://example.com/#minions` in my browser.

Is it possible to set any parameter to see `http://example.com/saltgui/#minions` in my browser?

Thank you.

### DIFF
--- a/saltgui/static/scripts/Router.js
+++ b/saltgui/static/scripts/Router.js
@@ -320,7 +320,7 @@ export class Router {
       }
       // push history state, so that the address bar holds the correct
       // deep-link; and so that we can use the back-button
-      let url = "/";
+      let url = config.NAV_URL ? config.NAV_URL : "/";
       let sep = "?";
       for (const key in pQuery) {
         const value = pQuery[key];
--- a/saltgui/static/scripts/panels/Panel.js
+++ b/saltgui/static/scripts/panels/Panel.js
@@ -885,8 +885,8 @@ export class Panel {
   static addPrefixImage (pElem, pImageName) {
     const img = document.createElement("img");
     const pngName = pImageName.replace(" ", "-").toLowerCase() + ".png";
-    img.setAttribute("src", config.NAV_URL + "/static/images/" + pngName);
-    img.setAttribute("onerror", "this.onerror=null; this.title='Unknown image, please report to SaltGUI team that image \\'" + pngName + "\\' is missing'; this.src='/static/images/UNKNOWN.png'");
+    img.setAttribute("src", "static/images/" + pngName);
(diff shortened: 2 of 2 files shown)


ddorosh:
Hi @erwindon 

thanks for your fast response.

It work with your changes in `/saltgui/static/scripts/Router.js b/saltgui/static/scripts/Router.js`.
When I set `NAV_URL` as `/saltgui/` I see `http://example.com/saltgui/#minions` in my browser. Exactly what I wanted. Thank you!


But I noticed one more thing. In spite of I have `API_URL` set to `/saltgui` my browser sends requests to `http://example.com/static` instead of `http://example.com/saltgui/static`

Untitled 1png

At the same time I have that js from my example at `http://example.com/saltgui/static`

Untitled 1png

Considering the proxy server it may cause a problem.

erwindon:
@ddorosh 
the includes for the `*.js` files are hardcoded in file `index.html`, the only html file that we have. but they are relative to `index.html`. because these are relative, they should not need API_URL or NAV_URL.

actually, I just made the URLs for the operating system relative too. variable NAV_URL is now only used to produce the updated URL in the address bar. this is a change in the branch from this PR. can you please pull that one in too and retry? please take an extra look at the OS images. before you login, use shift-refresh to make sure that you actually use the updated code. thx!

ddorosh:
@erwindon 
unfortunately it didn't help

Furthermore it "cancelled" previous commit and I see `http://example.com/#minions` again.

So what I did:
1. undo 8075c4f commit
2. created `saltgui` folder inside parent folder for SaltGUI
3. moved `static` folder into `saltgui` folder
file_treepng

4. in `index.html` file replaced `static/` with `saltgui/static/`

So now I have what I want `http://example.com/saltgui/#minions` in browser address bar and requests from browser to `http://example.com/saltgui/static/...`

Untitled2

Thank you for your help and your tips!

erwindon:
@ddorosh 
> Furthermore it "cancelled" previous commit and I see http://example.com/#minions again.

that is odd, as the additional commit was only about the handling of the OS images.
did you accidentality switch to another branch?

in theory, there should be no need to move the source code around when you want another base url. you should be able to instruct salt-api/rest_cherrypy to respond to different url prefixes. look in https://docs.saltproject.io/en/latest/ref/netapi/all/salt.netapi.rest_cherrypy.html for more details. (as you can determine from the wording, I do not have this working yet...)

I'll keep this PR open to investigate further.

ddorosh:
@erwindon 
I cloned it from saltguiurl branch. Hope it was correct.

I tried to play with salt-api/rest_cherrypy parameters, but I didn't get what I need

PR #480, opened 2022-10-15
show multimaster information

erwindon:
minions may be connected to multiple masters, but that is not visible in SaltGUI.

use `config.get` (or related functions) to get the minion configuration.
look at least for the following settings:
* master: the list of masters from the minion perspective
* random_master: true/false
* master_type: str/failover/func/disable

see also https://docs.saltproject.io/en/latest/ref/configuration/minion.html


erwindon:
archived

PR #483, opened 2022-10-17
Remove authentication methods 

erwindon:
It would be really nice to be able to remove unused authentication methods and/or set one as a default. Now it's always pam as default and the rest of the (unused) methods are in the pull down menu. What if I only use ldap?

I'd like to be able to remove unused authentication methods and set another one as default.

I'm not aware of alternatives but I'd be happy to try any if they are available.


erwindon:
a PR is now ready with the proposed solution:
* use file `salt-auth.txt` to set the wanted methods
* add  `CLEAR` to remove al standard methods and use only the once in the file (`CLEAR` can be on any line; but best to keep it at the top)
* with only one entry present, that becomes the default and the dropdown box is hidden
* with no entries present, you'll get `pam` for free and the previous rule applies, gives a warning on the console about it

still need to update the documentation.

can you please try it?

xzenor:
Works like a charm!

erwindon:
I added a few lines of documentation and then merged it.

erwindon:
thx for the improvement suggestion!

xzenor:
Anytime! And respect for how fast you pick this stuff up.. it's amazing.

PR #478, opened 2022-11-02
add support for parameter 'state_compress_ids'

erwindon:
Saltstack recently introduced parameter `state_compress_ids`.
This will change the output for some output models

Let SaltGUI adjust the output in the same way.

See also https://docs.saltproject.io/en/latest/ref/output/all/salt.output.highstate.html

sample output (`state_compress_ids=false`):
```
salt-minion-1:
  Name: apache-2 xyz - Function: cmd.run - Result: Changed Started: - 01:30:40.995510 Duration: 6.268 ms
  Name: apache-2 xyz - Function: cmd.run - Result: Failed Started: - 01:30:41.019488 Duration: 7.277 ms
  Name: apache-2 xyz - Function: cmd.run - Result: Changed Started: - 01:30:41.045495 Duration: 6.882 ms
  Name: apache-2 xyz - Function: cmd.run - Result: Failed Started: - 01:30:41.069218 Duration: 7.714 ms
```

sample output (`state_compress_ids=true`):
```
salt-minion-1:
  Name: apache-2 xyz (2) - Function: cmd.run - Result: Changed Started: - 01:35:59.463025 Duration: 26.798 ms
  Name: apache-2 xyz (2) - Function: cmd.run - Result: Failed Started: - 01:35:59.509360 Duration: 24.024 ms
```

note:
* count added to name part
* starttime is lowest of all starttimes
* duration is sum of all durations
* original saltstack fails when there is at least one error state; but we can skip those

### DIFF
--- a/saltgui/static/scripts/output/Output.js
+++ b/saltgui/static/scripts/output/Output.js
@@ -485,6 +485,11 @@ export class Output {
       summarySpan.append(span);
     }
 
+    const stateCompressIds = Utils.getStorageItem("session", "state_compress_ids", "false");
+    if (stateCompressIds === "true") {
+      summarySpan.append(Utils.createSpan("state-details-compressed", Character.NO_BREAK_SPACE + "(state details may be compressed)"));
+    }
+
     pMinionRow.append(summarySpan);
   }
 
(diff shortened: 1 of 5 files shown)


erwindon:
original:
afbeelding

compressed:
afbeelding

PR #462, opened 2022-11-20
in syndic: highstate page only shows the minions for which the master knows the keys

erwindon:
**Describe the bug**
highstate page only shows the minions for which the master knows the keys, minions behind salt-syndic are missing.

**To Reproduce**
1. Go to 'Highstate'
2. See that only directly-connected minions are shown

**Expected behaviour**
the page should show all minions


erwindon:
not a bug
to prevent old highstate reports to be shown, the highstate page is limited to the minions for which the keys are also known

PR #461, opened 2022-11-20
in syndic: for the minions that are behind the salt-syndic, click-row does not work

erwindon:
**Describe the bug**
in syndic: for the minions that are behind the salt-syndic, click-row does not work

**To Reproduce**
Use at least one minion that is behind a salt-syndic.
1. Go to 'grains'
2. Click on the row for the minion behind the salt-syndic
3. See that nothing happens

**Expected behaviour**
SaltGUI should zoom in on the grains-details of that minion.

**Additional context**
The menu-item for that row still works properly


erwindon:
W.I.P.

erwindon:
W.I.P.

erwindon:
syndic seems to be broken in the distribution
cannot support this now

PR #498, opened 2023-03-21
Limit Permission to specific Minions

erwindon:
Limiting SaltGUI access for different users to specific minions seems not to work.
Following the SaltGUI [permissions guide](https://github.com/erwindon/SaltGUI/blob/master/docs/PERMISSIONS.md) and saltstack [external authentication](https://docs.saltproject.io/en/latest/topics/eauth/index.html) docu, it should be possible to limit access for API.

Configure salt master with:
```
external_auth:
  pam:
    user1:
      - 'test*':
        - .*
        - '@runner'
        - '@wheel'
        - '@jobs'
```
Restart master and api.

SaltGUI only shows "(error)", no minions are listed.
When removing the test* line, everything works fine.

Salt master log shows the following lines, a token auth error seems to 
```
salt.auth         Compiled auth_list: [{'test*': ['.*', '@runner', '@wheel', '@jobs']}]
salt.utils.minions minions: {'test1', 'test2', 'prod1', 'prod2'}
salt.utils.minions Evaluating final compound matching expr: {'test1', 'test2'}
salt.master       Authentication failure of type "token" occurred.
```

Used Software:
salt 3004 on SUSE openleap 15.4, patched to latest.
SaltGUI 1.29.0

### DIFF
--- a/saltgui/static/scripts/Api.js
+++ b/saltgui/static/scripts/Api.js
@@ -94,7 +94,7 @@ export class API {
       params.tgt = pMinionId;
     } else {
       params["tgt_type"] = "glob";
-      params.tgt = "*";
+      params.tgt = Utils.getDefaultMinionTarget();
     }
     return this.apiRequest("POST", "/", params);
   }
@@ -110,7 +110,7 @@ export class API {
       params.tgt = pMinionId;
     } else {
       params["tgt_type"] = "glob";
-      params.tgt = "*";
+      params.tgt = Utils.getDefaultMinionTarget();
     }
     return this.apiRequest("POST", "/", params);
   }
@@ -125,7 +125,7 @@ export class API {
       params.tgt = pMinionId;
     } else {
       params["tgt_type"] = "glob";
-      params.tgt = "*";
+      params.tgt = Utils.getDefaultMinionTarget();
     }
     return this.apiRequest("POST", "/", params);
   }
@@ -140,7 +140,7 @@ export class API {
       params.tgt = pMinionId;
     } else {
       params["tgt_type"] = "glob";
-      params.tgt = "*";
+      params.tgt = Utils.getDefaultMinionTarget();
     }
     return this.apiRequest("POST", "/", params);
   }
@@ -155,7 +155,7 @@ export class API {
       params.tgt = pMinionId;
     } else {
       params["tgt_type"] = "glob";
-      params.tgt = "*";
+      params.tgt = Utils.getDefaultMinionTarget();
     }
     return this.apiRequest("POST", "/", params);
   }
@@ -171,7 +171,7 @@ export class API {
       params.tgt = pMinionId;
     } else {
       params["tgt_type"] = "glob";
-      params.tgt = "*";
+      params.tgt = Utils.getDefaultMinionTarget();
     }
     return this.apiRequest("POST", "/", params);
   }
@@ -180,7 +180,7 @@ export class API {
     const params = {
       "client": "local",
       "fun": "test.providers",
-      "tgt": "*"
+      "tgt": Utils.getDefaultMinionTarget()
     };
     return this.apiRequest("POST", "/", params);
   }
--- a/saltgui/static/scripts/Documentation.js
+++ b/saltgui/static/scripts/Documentation.js
@@ -94,7 +94,7 @@ export class Documentation {
     // the help text is taken from the first minion that answers
     // when no target is selected, just ask all minions
     if (target === "") {
-      target = "*";
+      target = Utils.getDefaultMinionTarget();
     }
 
     const commandField = document.getElementById("command");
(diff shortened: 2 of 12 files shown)


erwindon:
made this a pull-request

I've added a small function to inspect the limitation on the minion names.
all regular salt operations, that previously targeted `*`, now target the restricted range when applicable.

at this moment all minions are still shown, but details are only requested/presented for the restricted range.
the others continue to show "loading..."
I'll continue to work on this improve on that.
probably removing those minions from the table(s)...

@manfredw can you please try the version on this branch? (after fixing the `@` permissions)

manfredw:
> NOTE: the permission-list that you specified is not valid. the `@runner`/`@wheel`/`@jobs` clauses do not work when specified under a minion-pattern. on [external authentication](https://docs.saltproject.io/en/latest/topics/eauth/index.html), separate documentation parts deal with that.

Thanks for your quick reaction, I've tried your new branch and also changed configuration:
```
external_auth:
  pam:
    user1:
      - 'test*':
        - .*
      - '@runner'
      - '@wheel'
      - '@jobs'
```
Now it looks much better, but there are some other issues. The lists contains still all minions, but status stays on "loading" and access is denied (Error 401) when clicking on minions without permissions. Key management can't load any fingerprints, but accepts commands due to global @wheel permissions.

erwindon:
@manfredw 
thanks for looking at this.
I'll continue to compensate for the fact that not all data is available:
* remove the non-included minions from Minions/Grains/Schedules/Pillars/Beacons/Highstate
since SaltGUI cannot predict which minions are excluded, these screens will initially be populated with the names of all minions, the excluded minions will then be removed as the relevant answers come back from the salt-api.
* the Key management should be loading the fingerprints because these fall under the `@wheel` permissions, since it uses function `wheel.key.finger`
in the Keys screen, all minions will be visible because salt-api has no restrictions on the content of the wheel api
when you do not want that, you must exclude the `@wheel` permission
* (jobs)
in the Jobs screen, all jobs will be visible because salt-api has no restrictions on the content of the runner/jobs api
when you do not want that, you must exclude the `@jobs` and/or '@runner` permissions

erwindon:
@manfredw
ping...

manfredw:
Test of latest subset branch, hope that was the correct one:

Limited user:
The list of minions seems to be reduced, but there are still hosts shown it has no permissions.
Keys view is not showing fingerprint, shows full list of minions in state"loading..."
Recent Jobs view sometimes stays empty, showing "(error)", {}

Unlimited user:
Keys view is not showing fingerprint, "loading..."
Jobs-Highstate is showing "(error)", {}

General question:
How long is the GUI waiting for minions response? (timeout?)
What time window is used to get Jobs-Highstate? (minions with hourly highstate jobs seems to be OK, others with i.e. only daily highstates say "no job")

erwindon:
> The list of minions seems to be reduced, but there are still hosts shown it has no permissions.
The screen is initially filled with the full minion list, as wheel.
Do you use a single pattern to indicate the minions under `external_auth`?

> Limited user / Unlimited user
These are the same effect for both scenarios!
Anytime an `(error)` is shown, the tooltip on that text should show what the error is.
And also, the webbrowser console should give some useful information. Either on the text-console, or in the list of api calls.
Can you please report any error details that you see in these places?

> How long is the GUI waiting for minions response? (timeout?)
for that, SaltGUI relies entirely on the settings in the `master` file (which also contains the settings for `salt-api`).
e.g. setting `timeout`.

> What time window is used to get Jobs-Highstate? (minions with hourly highstate jobs seems to be OK, others with i.e. only daily highstates say "no job")
this question is unrelated to this issue.
it is easier to keep things separate.
can you please post this question as a separate issue?
just an empty one with the title "What time window is used to get Jobs-Highstate?" is ok.
I just want you to be the submitter of that issue.
we can discuss anything related to that page there.

> fingerprint
I did not notice that `wheel.key.finger` can be restricted to a minion pattern.
On my system, that did not cause problems, but we can do better here.
So, I'll make one more update to apply the filter directly in that request.
This is a contrast with `wheel.key.list_all` that only returns the whole list.

erwindon:
this is a topic that is very hard to get perfect.

the user may have multiple sets of minions configured in the `external_auth` setting.
for some sets, the permissions may be usable for SaltGUI, and for others it may be insufficient.
decision #1: SaltGUI will not target the minions separately for each permission-group. this leads to a lot of extra code, is hard to understand when compared to normal use-cases and should be very rare (even more rare than a simple permissions-subset).

the `external_auth` setting may have multiple sets of minions.
decision #2: SaltGUI should build a target that looks like `(group1) or (group2)`. if there is only a single group, then that group should be used directly (no parentheses added).

since the original requestor did not respond and we had no similar requests in the past:
decision #3: this change will be archived as `zzz-minionsets` and this PR closed without merge.

PR #504, opened 2023-04-11
Update Minions.js

AntonOvseenko:
Fix notification about upgrade or downgrade minion version

### DIFF
--- a/saltgui/static/scripts/panels/Minions.js
+++ b/saltgui/static/scripts/panels/Minions.js
@@ -666,13 +666,12 @@ export class MinionsPanel extends Panel {
         txt += MinionsPanel._addCveList("salt-minion", minionVersion, minionBugs, allCveKeys);
 
         if (outcome === "Minion requires update") {
-          txt += "\nThis salt-minion (" + minionVersion + ") is older than the salt-master (" + masterVersion + ")";
+          txt += "\nThis salt-minion (" + minionVersion + ") is older than the salt-master (" + masterVersion + ")\nUpgrade minion is highly recommended!";
         } else if (outcome === "Minion newer than master") {
-          txt += "\nThis salt-minion (" + minionVersion + ") is newer than the salt-master (" + masterVersion + ")";
+          txt += "\nThis salt-minion (" + minionVersion + ") is newer than the salt-master (" + masterVersion + ")\nDowngrade minion is highly recommended!";
         }
 
         if (txt) {
-          txt += "\nUpgrade is highly recommended!";
           if (allCveKeys.length > 0) {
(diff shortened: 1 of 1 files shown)


saltgui/static/scripts/panels/Minions.js
```diff
@@ -666,13 +666,12 @@ export class MinionsPanel extends Panel {
...
         if (outcome === "Minion requires update") {
-          txt += "\nThis salt-minion (" + minionVersion + ") is older than the salt-master (" + masterVersion + ")";
+          txt += "\nThis salt-minion (" + minionVersion + ") is older than the salt-master (" + masterVersion + ")\nUpgrade minion is highly recommended!";
         } else if (outcome === "Minion newer than master") {
-          txt += "\nThis salt-minion (" + minionVersion + ") is newer than the salt-master (" + masterVersion + ")";
+          txt += "\nThis salt-minion (" + minionVersion + ") is newer than the salt-master (" + masterVersion + ")\nDowngrade minion is highly recommended!";
         }
 
         if (txt) {
-          txt += "\nUpgrade is highly recommended!";
```
erwindon, line 675: removing this will no longer indicate that an upgrade is needed when the master+minions have an equal, but older, version for which CVEs are published

saltgui/static/scripts/panels/Minions.js
```diff
@@ -666,13 +666,12 @@ export class MinionsPanel extends Panel {
         txt += MinionsPanel._addCveList("salt-minion", minionVersion, minionBugs, allCveKeys);
 
         if (outcome === "Minion requires update") {
-          txt += "\nThis salt-minion (" + minionVersion + ") is older than the salt-master (" + masterVersion + ")";
+          txt += "\nThis salt-minion (" + minionVersion + ") is older than the salt-master (" + masterVersion + ")\nUpgrade minion is highly recommended!";
         } else if (outcome === "Minion newer than master") {
-          txt += "\nThis salt-minion (" + minionVersion + ") is newer than the salt-master (" + masterVersion + ")";
+          txt += "\nThis salt-minion (" + minionVersion + ") is newer than the salt-master (" + masterVersion + ")\nDowngrade minion is highly recommended!";
```
erwindon, line 671: what about "Upgrade master is highly recommended!"?

AntonOvseenko: It seems to me that you need to update the master to get new functionality and fix vulnerabilities. In this case, if the version of the master and the minion diverge, it is better to suggest downgrade the minion version

erwindon:
@AntonOvseenko
Sorry, I actually do not agree with the suggestion of a downgrade.
For this reason, I will not merge this PR.
Still, thanks for your participation!

PR #506, opened 2023-04-14
Update OutputHighstateTaskSaltGui.js

AntonOvseenko:
Example. Before changes id was: /data/saltgui id=Salt-Gui%20-%20download%20and%20extract%20archive (from projects/ati/salt-master/ubuntu.sls)
After changes: /data/saltgui id=[Salt-Gui - download and extract archive] (from projects/ati/salt-master/ubuntu.sls)


erwindon:
@AntonOvseenko 
great find!
that encoding action should not have been there.
I never noticed it because I personally stick to 'identifier' naming conventions.
I verified that all other calls to `encodeURIComponent` are in a proper context.

PR #507, opened 2023-04-14
Update Output.js

AntonOvseenko:
before: time tooltip ignore config parameter saltgui_datetime_fraction_digits. UTC time with 6 digits, local time with 0 digits after seconds
after: tooltip calculate time with this parameter. (UTC and local with saltgui_datetime_fraction_digits)

### DIFF
--- a/saltgui/static/scripts/output/Output.js
+++ b/saltgui/static/scripts/output/Output.js
@@ -199,7 +199,6 @@ export class Output {
     fractionSecondsPart = fractionSecondsPart.replace(/^.*[.]/, "");
     // remove everything after the digits
     fractionSecondsPart = fractionSecondsPart.replace(/[^0-9].*$/, "");
-    let originalFractionSecondsPart = fractionSecondsPart;
     // truncate digits to maximum length
     fractionSecondsPart = fractionSecondsPart.substring(0, dateTimeFractionDigits);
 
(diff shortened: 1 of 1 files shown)


erwindon:
@AntonOvseenko 
The tool-tip shows the original details by design.
Truncating to the number of digits specified by `saltgui_datetime_fraction_digits` happens in the regular HTML code to save screen-space and to enhance readability. The tool-tip reveals the the original value and additionally offers a local-time variant.

this is also mentioned in the documentation as:
> In all cases, a tooltip is added to a date+time field that shows the full representation of the date and time in both the local timezone and in UTC.

AntonOvseenko:
But in local time not full representation, only in seconds. In UTC is full. 
If the tooltip behaves according to the design, then sorry to bother you. image
I wanted to ask if there is an option in the config to enable or disable the tooltip? For example, when an option is selected in the `utc-local` or `local-utc` config, users might want to disable the tooltip

erwindon:
the intended behavior is that both representations show the full details.
as indicated by your screen-print, that behavior is broken, and I will fix that.

the answer to your question regarding tool-tips is that existing option `saltgui_tooltip_mode` can be set to:
* `simple` --> same tool-tip content, but using more light-weight technology, i.e. using the `title` attribute; and
* `none` --> prevents all tool-tips; and
* any other value --> causes the default behavior.

PR #515, opened 2023-05-24
Ability to display nodegroups

erwindon:
Hello! In /etc/salt/master added block with nodegroups:

Is it possible to display nodegroups ("sandbox") in SaltGUI on the page with minions?

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -30,6 +30,7 @@ such as `yubico_client`, or execution modules such as `boto3_sns`.
 - View the schedules for a particular minion
 - View the values for pillars for a particular minion
 - View the beacons for a particular minion
+- View the nodegroups with their member minions
 - View the live events on the salt-event bus
 - View internal documentation for any salt command
 - View external documentation for any salt command
@@ -269,6 +270,14 @@ e.g.:
 saltgui_public_pillars:
     - pub_.*
 ```
+
+## Nodegroups
+The Nodegroups page shows all minions, but groups the minions by their nodegroup.
+Since the group membership can only (reliably) be determined by executing salt commands, the overview takes some time to build.
+When a minion happens to be a member of multiple nodegroups, then a row for that minion will appear mutiple times in the overview.
+When using targeting based on nodegroups, the salt-system does not filter out the non-accepted minions.
+Therefore also the rejected, denied and unaccepted minions will show up.
+Even the minion names that are in the nodegroup, but which do not actually exist, will show up as 'unknown'.
 
 ## Highstate
 The Highstate page provides an overview of the minions and their latest state information.
--- a/docs/README.md
+++ b/docs/README.md
@@ -33,6 +33,7 @@ such as `yubico_client`, or execution modules such as `boto3_sns`.
 - View the live events on the salt-event bus
 - View internal documentation for any salt command
 - View external documentation for any salt command
+- View minions organized by node-group
 - Define your own custom documentation for commands
 - Match list of minions against reference list
 - Match status of minions against reference list
--- a/saltgui/static/scripts/Api.js
+++ b/saltgui/static/scripts/Api.js
@@ -177,17 +177,22 @@ export class API {
     return this.apiRequest("POST", "/", params);
   }
 
-  getLocalTestProviders (pNodegroup) {
+  getLocalTestProviders () {
     const params = {
       "client": "local",
-      "fun": "test.providers"
+      "fun": "test.providers",
+      "tgt": "*"
     };
-    if (pNodegroup) {
-      params["tgt"] = "N@" + pNodegroup;
-      params["tgt_type"] = "compound";
-    } else {
-      params["tgt"] = "*";
-    }
+    return this.apiRequest("POST", "/", params);
+  }
+
+  getLocalTestVersion (pNodegroup) {
+    const params = {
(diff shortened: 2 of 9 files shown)


erwindon:
a first part of the implementation is done:
* added a new page "nodegroups" with the corresponding menu item; and
* the page shows a heading for each of the nodegroups and a heading for any minion that is not part of any nodegroup.
feel free to look at the intermediate results on branch `nodegroups`.

the information displayed for the minions is exactly like on the "minions" page (the default home page), EXCEPT:
* the "nodegroups" page does not use (wheel) `minions.connected` to provide additional info on minion connectivity; and
* the "nodegroups" page does not use (runner) `manage.versions` to provide additional info on minion versions. this info indicates which minions (or even the master) are either inconsistent or older than the latest available.

next steps are at least:
* asynchronously determine the membership per nodegroup using a simple `test.ping` operation. even when the ping fails, we know that a minion is part of that group. for any minion that is a known nodegroup-member, it is taken away from the "not in any nodegroup" section and placed in the section for the nodegroup. if any ping operation shows that a minion is _also_ part of another group, then the row will be duplicated for that; and
* when "not in any nodegroup" becomes empty, that heading is removed. but other empty nodegroups just stay when they have no minions; and
* add a drop-down menu's to each heading for the nodegroups, includes the special case for "not in any nodegroup" which will be like `(not N:grp1 and not N:grp2 and not N:grp3)`; and
* add the minion count to each heading; and
* update the status line to show the number of minions and the number of (real) nodegroups; and
* update more of the documentation.

the whole implementation takes several days.

erwindon:
> add a drop-down menu to each headings for the nodegroups, includes the special case for "not in any nodegroup" which will need something like `not N:grp1 and not N:grp2 and not N:grp3`

added that

erwindon:
> asynchronously determine the membership per nodegroup using a simple test.ping operation. even when the ping fails, we know that a minion is part of that group. for any minion that is a known nodegroup-member, it is taken away from the "not in any nodegroup" section and placed in the section for the nodegroup.

added that. looks a bit jumpy, but for now that is accepted.

> if any ping operation shows that a minion is also part of another group, then the row will be duplicated for that

extra work here. duplicating an html-element (like a table row) loses all handlers that are part of that. i.e. clicking on the ip-number, the row itself or in the dropdown-menu of the row is now broken...

erwindon:
@dzhumshoot 
the new page is ready for testing.
can you give it a try?
please use `git pull -r` to update the git branch.

erwindon:
> But if I navigate between tabs and revisit the page, nothing happens, there is no play/pause button. I need to refresh the page in browser to make it work.
thx!
organizing the minions in nodegroups is a backgroundtask, which may have many iterations. therefor a play/pause button is present on this page and likewise for technically similar pages.
but the "play/pause" button is never reset here, therefore it remains in its "pause" setting.
should be an easy fix, and I'll validate the other screens as well.

> layout
I noticed a layout problem for the offline minions in the screenprints.
I've created a separate issue #519 for that, as that is an existing problem.

erwindon:
additionally, now using an `—` (em-dash) instead of `---`

erwindon:
@dzhumshoot 
can you please try again?
please report any error, but also things that are still missing, or improvements, etc.

erwindon:
actually, the layout problem does not exist on other pages.
therefore I'll fix it here

erwindon:
fixed a logic problem in the handling of offline-minions (still need to fix the layout problem)

erwindon:
fixed the layout problem

erwindon:
@dzhumshoot
can you please try again?
please report any error, but also things that are still missing, or improvements, etc.

dzhumshoot:
Now the page is updated correctly, including when moving through the pages without the need to refresh the page in the browser, thanks!
But problem with offline minions is still here.



Is it possible to display also the number of connected and offline minions in each nodegroup?

erwindon:
> But problem with offline minions is still here.
I forgot the `git push` part :-)
done that now

erwindon:
> Is it possible to display also the number of connected and offline minions in each nodegroup?

yes, I'll add that

erwindon:
and I'm thinking of excluding the nodegroup-headers from the text-search logic so that these are never hidden...

erwindon:
> But problem with offline minions is still here.

this is fixed

> Is it possible to display also the number of connected and offline minions in each nodegroup?

this is added

> [...]excluding the nodegroup-headers from the text-search logic so that these are never hidden when searching[...]

this is added

@dzhumshoot
can you please try one more time?

erwindon:
just as I thought the work was finished...
targeting a node group includes all minions in nodegroup-definition, whether their key is accepted or not.
this happens at least when the nodegroup consists of a simple list of minion-ids.
this causes an operation on a non-existing row because those minions were not initially in the list, as we only include accepted minions there.
I'll compensate for that...

erwindon:
solved it:
the nodegroups panel now handles unaccepted/denied/rejected and unknown minions as members of a nodegroup.

dzhumshoot:
I checked the latest version - everything seems to be working correctly, thanks!

erwindon:
@dzhumshoot 
did you see my request in #517 ?
(github is not always cooperative when you one mentions a so-far non-participating user)

PR #520, opened 2023-06-09
Readme may be incorrect for grains

erwindon:
current readme says 
saltgui_preview_grains:
    - "osrelease_info"

single and double quotes both did not work.

commit #166 where that was adding has no quotes. which worked for me.

saltgui_preview_grains:
    - osrelease_info

may want to update so others dont bang their head as much :)

### DIFF
--- a/saltgui/static/scripts/panels/Grains.js
+++ b/saltgui/static/scripts/panels/Grains.js
@@ -20,22 +20,25 @@ export class GrainsPanel extends Panel {
     this.addTable(["Minion", "Status", "Salt version", "OS version", "Grains", "-menu-"]);
     this.setTableClickable();
 
-    // collect the list of displayed extra grains
-    this.previewGrains = Utils.getStorageItemList("session", "preview_grains");
-
-    // add the preview columns (before we sort the table)
-    // the div is not added to the DOM yet
-    const tr = this.div.querySelector("#grains-table-thead-tr");
-    for (const previewGrain of this.previewGrains) {
-      const th = Utils.createElem("th", "", previewGrain);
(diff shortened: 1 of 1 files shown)


erwindon:
converted the issue to a PR

erwindon:
I've added the fix to this PR.

@sharkmerry7 
can you pull this PR (branch `preview-grains`) and test it?
the preview-grains should be visible when the Grains page is first shown and no refresh actions should be needed.

sharkmerry7:
seems to be same behavior

all i did was:

git fetch
git switch preview-grains
restart salt-api and salt-master

do i need to do something else?

sharkmerry7:
seems to be same behavior

i did 
git fetch
git pull origin master
git switch preview-grain
restart salt-api and salt-master

do i need anotehr step?

erwindon:
> restart salt-api and salt-master
that is not needed :-)

the branch-name actually has an extra `s` at the end.

your browser may use the old version of that page from its own cache
typically using SHIFT-F5 forces the page to be reloaded and all files to be reloaded from disk

sharkmerry7:
I mistyped branch name in my comment. had it right on the host. 

shift+F5 did it! Thank you!!

erwindon:
and now the big question...
does the grains-preview work as expected now?

sharkmerry7:
shift+F5 did it! Thank you!!

erwindon:
merged it and spotted one more thing...
the creation of the columns was moved to a better location.
but sorting was still setup for the initial set of columns.
that no longer included the columns for the previewed grains.
so moved one more line of code to fix the fix.

PR #523, opened 2023-06-10
collect_stats=true always causes a forced logout

erwindon:
**Describe the bug**
A bug in SaltStack itself...
When switching on `collect_stats` for salt-api, the call `/stats` fails with an internal server error response (code 500).
this happens at least for salt-versions up to 3006.1, which is the lasted at this moment. 
SaltGUI logs-out on any error from this call.

**Expected behaviour**
SaltGUI should ignore this specific error since it does not actually provide information on the login status.


erwindon:
fixed by inspecting the actual error code
with 500, ignore and stop /stats timer
otherwise still logout

PR #533, opened 2023-07-15
prevent problems with long job argument list

erwindon:
some salt command have long arguments lists.

the full argument list was used as the title of the Job page.
this should be shortened to the first 50 characters only.

but when that title is shortened, there is no more insight in the argument list of the command.
therefore, in that case, add the full command at the top of the output panel.


erwindon:
result:
afbeelding

erwindon:
@tazaki fyi

tazaki:
Confirmed this fixes my issue after the patch to app.py

PR #536, opened 2023-08-11
Switch default runtype from "normal" to "async"

spankywetfish:
In general use with multiple hosts it is (IMHO) more practical to have the jobs run asynchronously so that there is no need to wait for completion before performing the next task.

The changes made to RunType.js reflect that priority and default the job input to "async".  
I've been running with this configuration for approximately 18 months with no known issues.

N.B.
Apologies if this is not the correct format for a pull request, it's a first for me.


erwindon:
@spankywetfish
Technically, the PR is fine.
But I cannot change the default behavior of a feature, unless it is an improvement for everyone.
I cannot accept this as is.
But all is not lost...
There are alternatives that should work for all:
1. make the initial value based on a variable in the `/etc/salt/master` file
rather technical and it would apply to all users
1. make the initial value based on a variable in some `*.txt` file
rather technical and it would apply to all users
1. use 2 "Run command" buttons, one "Run command normal" and also "Run command async"
no configuration, is per-use, but complicates GUI and conflicts with possible extensions in the future (e.g. batched runs).
1. make the initial value based on a plain cookie, set according to the latest use
no configuration, is per-user/per-browser, still looks simple, must make the default more visible to prevent confusion. information leak through use of permanent cookies is negligible.

how about number 4?

erwindon:
Additionally, I've created an implementation of proposal 4. See branch [runtype4](https://github.com/erwindon/SaltGUI/tree/runtype4).

spankywetfish:
It looks like we've both been doing the same thing slightly different ways, I've also updated Api,js  so that the runtype is persistent across sessions. I'd hazzard that as I consider myself in no way a programmer, yours implementation is more elegant.

saltgui/static/scripts/RunType.js
```diff
@@ -21,10 +21,10 @@ export class RunType {
 
     switch (runType) {
     case "normal":
-      // now that the menu is used show the menu title
       RunType.menuRunType.setTitle("Normal");
       break;
     case "async":
+      // now that the menu is used show the menu title
```
erwindon, line 27: no need to move the comment to here as the default remains "normal".

saltgui/static/scripts/Api.js
```diff
@@ -41,10 +41,12 @@ export class API {
   }
 
   static _cleanStorage () {
-    // clear local storage except key 'eauth'
+    // clear local storage except key 'eauth' and 'runtype'
     const eauth = Utils.getStorageItem("local", "eauth");
+    const runtype = Utils.getStorageItem("local", "runtype");
     Utils.clearStorage("local");
     Utils.setStorageItem("local", "eauth", eauth);
+    Utils.setStorageItem("local", "runtype", runtype);
```
erwindon, line 50: actually, I completely forgot that we erase most local storage, but this handles that part nicely

saltgui/static/scripts/RunType.js
```diff
@@ -55,6 +60,8 @@ export class RunType {
     if (runType === undefined || runType === "") {
       return "normal";
```
erwindon, line 61: this should now be updated to `runType = "normal";`

saltgui/static/scripts/RunType.js
```diff
@@ -43,7 +43,12 @@ export class RunType {
   }
 
   static setRunTypeDefault () {
-    RunType.menuRunType._value = "normal";
+    // Retrieve last used runType
+    RunType.menuRunType._value = Utils.getStorageItem("local", "runtype");
+    // Set default if previous runtype not stored
+    if (RunType.menuRunType._value != "normal" && RunType.menuRunType._value != "async") {
+      RunType.menuRunType._value = "normal";
+    }
```
erwindon, line 51: that's correct, but usually any logic is done using local variables.
then when the correct value is there, assign to the intended object property
(please use with the original comments)
```
let runType = Utils.getStorageItem("local", "runtype");
if (runType !== "normal" && runType !== "async") {
    runType = "normal";
}
RunType.menuRunType._value = runType;
``` 

note also that "!==' is preferred instead of "!="
please run "npm run eslint" before committing
(and then also find some trailing spaces...)

erwindon:
I've hidden my branch `runtype4`.
we're nearly there, please look at the comments with the js code...

saltgui/static/scripts/RunType.js
spankywetfish: That makes sense and makes sure the initial runtype is saved due to no longer returning from the function prior.
Changed runType from `const` for obvious reasons.

saltgui/static/scripts/RunType.js
spankywetfish: Makes sense. define once use many.

Fixed and linted.

saltgui/static/scripts/RunType.js
```diff
@@ -43,18 +43,25 @@ export class RunType {
...
   static getRunType () {
-    const runType = RunType.menuRunType._value;
+    let runType = RunType.menuRunType._value;
     if (runType === undefined || runType === "") {
-      return "normal";
+      runType = "normal";
     }
+    // Store last used runType
+    Utils.setStorageItem("local", "runtype", runType);
     return runType;
```
erwindon, line 65: when a page-refresh is done, all GUI values are unavoidably lost.
this function would then reset the value to "normal" and also save it.
- saving the value should only be done in function _updateRunTypeText, as that is the callback function from the dropdownlist.
- instead of `runType = "normal"`, the saved value should be retrieved: `runType = Utils.getStorageItem("local", "runtype", "normal");` (which itself will have a default of 'normal')

spankywetfish: Moved to ` _updateRunTypeText,` and suggested modification for retrieval applied, all tested and linted.

erwindon:
for a next time...
please use a dedicated branch, which you can re-base until it is merged.
I rolled-back this merge and made it a regular commit.

erwindon:
thanks for the addition!

PR #545, opened 2023-10-31
feature request: keyboard shortcuts, ESC key

erwindon:
In addition to clicking to [X] on the Manual Run modal window, it would be great to have the same effect by pressing the ESC key. Also, on the main page, some keyboard shortcuts could make tasks easier. For example:

- "r" for opening the Manual Run modal
- "m" for minions
- ...and so on.
- "t" for templates. on this page, "1", "2", "3" etc. could apply the relevant template

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -38,6 +38,8 @@ such as `yubico_client`, or execution modules such as `boto3_sns`.
 - Define your own custom documentation for commands
 - Match list of minions against reference list
 - Match status of minions against reference list
+- Keyboard control for top-level navigation
+- Keyboard control to apply templates
 
 
 ## Quick start using PAM as authentication method
--- a/docs/README.md
+++ b/docs/README.md
@@ -184,6 +184,7 @@ saltgui_templates:
         description: First template
         target: "*"
         command: test.fib num=10
+        key: "f"
     template2:
         description: Second template
         targettype: glob
@@ -212,6 +213,9 @@ saltgui_templates:
 When at least one template is assigned to a category, then you can select a template category before
 selecting the actual category. Otherwise that choice remains hidden. Templates can be in multiple categories
 when a list of categories is assigned.
+
+When at least one template is assigned to a key, then you can select a template by typing that key
+while using any other screen in SaltGUI. Otherwise that choice remains hidden.
 
 
 ## Jobs
--- a/saltgui/index.html
+++ b/saltgui/index.html
@@ -24,10 +24,12 @@
   <body>
     <header id="header" class="no-print">
       <h1 class='logo'><div id="logo">SaltGUI</div></h1>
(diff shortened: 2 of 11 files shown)


erwindon:
turned issue into PR

erwindon:
first change is a general improvement.
hiding a useless column had some issues, that is now fixed.
first (and so far only) column to hide was the "Category" column for the templates overview

erwindon:
second change is to allow assignment of a key to a template.
any regular key can be bound to a template by adding the property "key" to the original template definition, e.g. `key: "7"`.
note that alphabetic keys will be used in the near future to select regular menu-items (to switch to a specific page).
pressing the key will open the command-box and fill it with the values from the associated template.
this can be initiated from any page as long a no other input-control has the focus (should only happen with the 2 fields of the CommandBox and the radio-buttons on the Options page)

erwindon:
feel free to install the software from the branch and try it...

ZsBT:
I've tested it thoroughly and did not found any issues. It works from any page as expected, even on the Manual Run modal window ;)
Maybe a hint on the Manual Run hamburger menu could be useful, but it's not important as shortcut keys can be seen on the Templates page.

manualrun-template

I am thinking of something like Reload HAproxies [3] , Reload Nagios [7] , in my case.

All in all, it works great, thanks for the implementation.

erwindon:
> I've tested it thoroughly and did not found any issues.

thx!

> It works from any page as expected, even on the Manual Run modal window ;)

I noticed that too, but found no reason to take the extra effort to prevent it.

> Maybe a hint on the Manual Run hamburger menu could be useful, but it's not important as shortcut keys can be seen on the Templates page.

good point. can do that.

> [menu-bar keys]

will be working on that after the previous one.

erwindon:
just added the display of the shortcut in the templates menu.

erwindon:
added page navigation by using first character of function-name.
"c" opens the command-box.
now working on making that more visible.

erwindon:
added highlight of first menu-item character in the main menu
works completely in the wide-screen edition of the main menu
only works for the major menu-items in the small-screen edition of the main menu
also nothing for the `>_` button yet

ZsBT:
I've just tried this branch. For me, the navigation is lightning-fast, I could not find any bugs.

erwindon:
thx again for testing!

> only works for the major menu-items in the small-screen edition of the main menu

that is the only remaining work for now: to get the keyboard shortcut working for all main-menu-items in the small-screen edition...

erwindon:
I've added the code to show the keyboard indicators.
this concludes the set of changes.
can you please do one final test?
note that I've used `git push --force` on code that you already have to fix a cosmetic issue. I suggest `git pull -r` to update.

ZsBT:
Tested again and all works as expected. Thanks!

Issue #538, opened 2023-08-22
Cache minion data to store grains data  and last online status for offline minions


Bearbobs:
**Is your feature request related to a problem? Please describe.**
It's hard to analyze the infra if the minions are offline so having a cache layer will be helpful to save the data till the minion is offline

**Describe the solution you'd like**
Having a cache connection with a database to save the status of offline minions

Bearbobs:
without persistent storage it will be pointless I guess as the browser storage will still be dependent on the minion's last online status for that browser.
The workflow which I thought of looks like this:
1.) A backend agent to ping salt-master on schedule to save all minion info in a database
2.) API from this agent to serve cache mode data on saltgui and a toggle to switch between cache and live mode.
I will keep you posted for upstream contribution

erwindon:
@mbgevers looks promising. can you please open a new issue for this? just repeat the question there. if I copy your question, then I become the owner instead of you.... let's have all further discussion in the new issue.

PR #551, opened 2023-12-20
use cache.grains function to display inventory data

erwindon:
Hi,

When doing some research on why i never get cached minion data on SaltGUI i discovered https://github.com/erwindon/SaltGUI/issues/538 which mentions there is a need for a database, however i found another possible solution:

Instead of using a Backend-Database to display some inventory/stored grains data of minions it should be possible to use the cache.grains function as documented in https://docs.saltproject.io/en/latest/topics/cache/index.html it looks like this cache also supports multiple backends/modules, see https://docs.saltproject.io/en/latest/ref/cache/all/index.html#all-salt-cache for a list so the database question is also answered.

It looks like implementing this could be really easy, actually a config-file option which exchanges the "normal grains command" with the "cached grains command" should do the job perfectly. After that only the documentation of salt for cached grains has to be followed.

If somebody wants to implement, just ensure to not hit https://github.com/saltstack/salt/issues/48694

As we are just starting with salt, i would be willing to test.

Kind regards

Mike

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -40,6 +40,7 @@ such as `yubico_client`, or execution modules such as `boto3_sns`.
 - Match status of minions against reference list
 - Keyboard control for top-level navigation
 - Keyboard control to apply templates
+- Choose between live info and cached info for grains/pillar
 
 
 ## Quick start using PAM as authentication method
@@ -276,6 +277,10 @@ saltgui_ipnumber_prefix:
 ```
 The display of the IP-numbers can simply be disabled by choosing a non-existing grain or by choosing a non-existing prefix.
 
+SaltGUI will retrieve cached grains information when variable `saltgui_use_cache_for_grains` is set to `true`.
+In that case, unreachable minions will appear without warnings for that.
+In all cases, the information may be less accurate.
+
 ## Pillars
 Pillars potentially contain security senstitive information.
 Therefore their values are initially hidden.
@@ -289,6 +294,10 @@ e.g.:
 saltgui_public_pillars:
     - pub_.*
 ```
+
+SaltGUI will retrieve cached pillar information when variable `saltgui_use_cache_for_pillar` is set to `true`.
+In that case, unreachable minions will appear without warnings for that.
+In all cases, the information may be less accurate.
 
 ## Nodegroups
 The Nodegroups page shows all minions, but groups the minions by their nodegroup.
--- a/docs/README.md
+++ b/docs/README.md
@@ -280,6 +280,7 @@ The display of the IP-numbers can simply be disabled by choosing a non-existing 
 SaltGUI will retrieve cached grains information when variable `saltgui_use_cache_for_grains` is set to `true`.
 In that case, unreachable minions will appear without warnings for that.
 In all cases, the information may be less accurate.
+A warning for offline minions is only shown on the Minions panel.
 
 ## Pillars
 Pillars potentially contain security senstitive information.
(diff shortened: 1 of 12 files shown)


erwindon:
changed this github issue to a PR

erwindon:
decision: let's not bother with the `mine` function. that is on an experimental branch and still incomplete

erwindon:
@mbgevers 
The branch now contains a working version of SaltGUI that is capable of using cached grains and/or cached pillar information.
The current behavior remains the default.
But that can be overruled by adding one or both of these settings to the `/etc/salt/master` file:
```
saltgui_use_cache_for_grains: true
saltgui_use_cache_for_pillar: true
```
but you can also switch it on for the current session in the Options screen (ctrl-click on the SaltGUI logo).
On all screens where it applies, a warning message is added about the use of cached information.

Can you give it a try?
Let me know any problems or suggestions...

mbgevers:
> @mbgevers The branch now contains a working version of SaltGUI that is capable of using cached grains and/or cached pillar information. The current behavior remains the default. But that can be overruled by adding one or both of these settings to the `/etc/salt/master` file:
> 
> ```
> saltgui_use_cache_for_grains: true
> saltgui_use_cache_for_pillar: true
> ```
> 
> but you can also switch it on for the current session in the Options screen (ctrl-click on the SaltGUI logo). On all screens where it applies, a warning message is added about the use of cached information.
> 
> Can you give it a try? Let me know any problems or suggestions...

Only seen this comment now, yeah for sure will give it a try

mbgevers:
Hi, doing some testing at the moment, looks good so far, however found some drawbacks. There should be some symbol indicating if a minion is online (thus no cache used) or offline (cache used) what do you think about that? So cached should be a fallback. Thx much for your work already!

erwindon:
> [...] if a minion is online (thus no cache used) or offline (cache used) [...]
reminder: technically, SaltGUI is now either using cached information or live information, never a combination of that

> There should be some symbol indicating if a minion is online [...] or offline [...]
previously that was visible by the indication "offline" in the status column.
in the no-cache case, that remained visible as no grains were returned from offline minions.

I'll investigate on how to display the offline status in a different way.

mbgevers:
> > [...] if a minion is online (thus no cache used) or offline (cache used) [...]
> > reminder: technically, SaltGUI is now either using cached information or live information, never a combination of that
> 
> > There should be some symbol indicating if a minion is online [...] or offline [...]
> > previously that was visible by the indication "offline" in the status column.
> > in the no-cache case, that remained visible as no grains were returned from offline minions.
> 
> I'll investigate on how to display the offline status in a different way.

I just had the idea, it should be sufficient if you just write the "cache timestamp" or "last cache update time" if that info is somehow available. Another maybe even easier method would be to only load cache if device is offline.

erwindon:
> [...] it should be sufficient if you just write the "cache timestamp" or "last cache update time" if that info is somehow available.
no such facility is described in https://docs.saltproject.io/en/latest/ref/runners/all/salt.runners.cache.html

> Another maybe even easier method would be to only load cache if device is offline.
there should be no problem to always using cached information. it is used often in SaltStack. e.g. each time you the `salt` command with a grain-based target, the cache will be used to determine the actual target. see also https://docs.saltproject.io/en/latest/topics/targeting/grains.html.

The problem that we still need to solve is that regular salt-functions (those executed on minions) indirectly provide availability-indicator as extra information because they fail. In SaltGUI this was even optimized by using an extra call (wheel.minions.connected), so that the availability information is already visible while you are waiting. But with the use of cached information, that indicator is overwritten by the cached information. I'm now changing the presentation of that, so that it remains visible in all cases.

erwindon:
@mbgevers 
I've added a visual indication on the main screen when a minion is offline.
does that work for you too?

mbgevers:
> @mbgevers I've added a visual indication on the main screen when a minion is offline. it is added only for minions for which there is information available, therefore the traditional case is unchanged. does that work for you too?

That looks nice, thank you.

PR #552, opened 2023-12-22
add some more comments on Cached Grains

mbgevers:
add some more comments on Cached Grains

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -277,9 +277,10 @@ saltgui_ipnumber_prefix:
 ```
 The display of the IP-numbers can simply be disabled by choosing a non-existing grain or by choosing a non-existing prefix.
 
+### Cached Grains
 SaltGUI will retrieve cached grains information when variable `saltgui_use_cache_for_grains` is set to `true`.
 In that case, unreachable minions will appear without warnings for that.
-In all cases, the information may be less accurate.
+In all cases, the information may be less accurate, however the Information may help to create a basic inventory, as by nature an Inventory is never absolutely up-to-date till its next inventory update run.
 
 ## Pillars
 Pillars potentially contain security senstitive information.
--- a/docs/README.md
+++ b/docs/README.md
@@ -40,6 +40,7 @@ such as `yubico_client`, or execution modules such as `boto3_sns`.
 - Match status of minions against reference list
 - Keyboard control for top-level navigation
 - Keyboard control to apply templates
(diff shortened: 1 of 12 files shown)


docs/README.md
```diff
@@ -277,9 +277,10 @@ saltgui_ipnumber_prefix:
 ```
 The display of the IP-numbers can simply be disabled by choosing a non-existing grain or by choosing a non-existing prefix.
 
+### Cached Grains
```
erwindon, line 280: the rest of this chapter is not divided in 3rd-level sections. it either should be left out, or a section "### Regular Grains" should be added just below "## Grains"

erwindon, line 280: the same structure should be applied below "## Pillars"

docs/README.md
```diff
@@ -277,9 +277,10 @@ saltgui_ipnumber_prefix:
...
 SaltGUI will retrieve cached grains information when variable `saltgui_use_cache_for_grains` is set to `true`.
 In that case, unreachable minions will appear without warnings for that.
-In all cases, the information may be less accurate.
+In all cases, the information may be less accurate, however the Information may help to create a basic inventory, as by nature an Inventory is never absolutely up-to-date till its next inventory update run.
```
erwindon, line 283: Information and Inventory should be without capital, do you have a German language background? :-)

erwindon, line 283: should this explanation also be added to the Pillars section?

erwindon, line 283: ", however the"
better to actually start a new sentence.
therefore:
". However, the"

erwindon, line 283: the whole explanation is a bit unclear for me.
SaltGUI does not contribute to building this inventory as that inventory was already maintained by the salt-master. This change just uses that information.

erwindon (changes requested):

erwindon:
@mbgevers 
github closed this PR automatically because the base branch no longer exists.
feel free to re-open it... (after rebasing it to `main`)

docs/README.md
mbgevers: Hi,
lol yes i have, but actually Swiss not German.

mbgevers: I have to read some more about Pillars, as i don't understand them very well yet, then of course i may add that.

mbgevers: Will work on improvement here

mbgevers: Sure, has to be improved

PR #570, opened 2024-04-15
add "saltgui_max_show_highstates" and "saltgui_max_highstate_states" in config options

AntonOvseenko:
With a large number of salt minions, it is not enough to display only the last 10 highstates. I suggest leaving the default value 10 but adding the ability to change it through the config. I would also like to be able to customize the number of states for detailed display (default is 20)

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -310,7 +310,9 @@ Even the minion names that are in the nodegroup, but which do not actually exist
 
 ## Highstate
 The Highstate page provides an overview of the minions and their latest state information.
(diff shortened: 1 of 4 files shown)


erwindon:
> With a large number of salt minions, it is not enough to display only the last 10 highstates.
That should be unrelated... Unless of course you are running the highstate jobs individually or in small groups.

Anyway, you provide enough reason to make it a config-variable.
I'll look through the details but the overall impression is that this PR is ok.

erwindon:
could not find a single thing to complain about :-)

erwindon:
THX!

PR #571, opened 2024-04-17
change color for changed tasks (before not different from not-changed)

AntonOvseenko:
image

I suggest changing the task-changed display color for better readability of states

PR #574, opened 2024-04-25
Open a row/link in a new tab

erwindon:

Many pages like pillars, grains, jobs, etc., contain an extensive list that might take a while to load, depending on the number of minions. In this context, whenever you click in the row, it loads the content on the same page. But once you go back, it needs to load everything again.

There is a good side to always getting the latest data. But it is very annoying when you have a huge list or a filter to load, and when you get back to the page, it loads the default content again. It would be amazing to open a new tab for each job we want to look deeper into without the need to load the whole list for each of them (that would also decrease the load on the server calls).

Create a configuration to force every row click to open in a new tab so the user can decide whether this behavior is more convenient. Right-clicking and opening a new tab could also be possible as a solution.

Adding an extra button on this kind of page to open the content in a new tab _(if the row behavior is more challenging to implement)_.

**Additional context**

I could contribute if you gave me some guidance. I saw panels and page folders, but I'm not sure exactly how to add things there.

Thanks in advance for reading the message and maintaining this great project. Cheers!

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -34,6 +34,7 @@ such as `yubico_client`, or execution modules such as `boto3_sns`.
 - View internal documentation for any salt command
 - View external documentation for any salt command
 - View minions organized by node-group
+- View details optionally in a separate windows
 - Define your own custom documentation for commands
 - Match list of minions against reference list
 - Match status of minions against reference list
@@ -114,6 +115,16 @@ Note that SaltGUI cannot inspect the `master` file to see which authentication m
 This is because the salt-api can only read the file after login.
 
 See the [EAUTH documentation](https://docs.saltstack.com/en/latest/topics/eauth/index.html) and the [Salt auth source code](https://github.com/saltstack/salt/tree/master/salt/auth) for more information.
+
+
+## Browser tabs
+SaltGUI is a single page web-application.
+In cases where you would zoom in on details, it is possible to open a new browser tab with the requested details.
+Use CTRL-click to open the page in a new tab.
+Use ALT-click to open the page in a new tab and also make that the current tab.
+This works for clicks on table-rows and for clicks on popup-menu items.
+These functions are a bit browser-dependent, but all major browsers seem to follow this behavior.
+When a new tab is opened by SaltGUI, it does not contain the menu bar items, secondary panels or a close-button inside the page.
 
 
 ## Command Box
--- a/saltgui/static/scripts/Router.js
+++ b/saltgui/static/scripts/Router.js
@@ -278,6 +278,18 @@ export class Router {
     }
   }
 
+  static _cancelSelections () {
+    // see https://stackoverflow.com/questions/3169786/clear-text-selection-with-javascript
+    const sel = window.getSelection ? window.getSelection() : document.selection;
+    if (sel) {
+      if (sel.removeAllRanges) {
+        sel.removeAllRanges();
+      } else if (sel.empty) {
+        sel.empty();
+      }
+    }
+  }
+
   static updateMainMenu () {
     const pages = Router._getPagesList();
 
@@ -370,6 +382,7 @@ export class Router {
         // forward navigation
         if (inNewWindow) {
           // in a new window
+          Router._cancelSelections();
           window.open(url);
           return;
         }
--- a/saltgui/static/scripts/Router.js
+++ b/saltgui/static/scripts/Router.js
@@ -300,7 +300,7 @@ export class Router {
   // pForward = 0 --> normal navigation
   // pForward = 1 --> back navigation using regular gui
   // pForward = 2 --> back navigation using browser
-  goTo (pHash, pQuery = {}, pForward = 0) {
+  goTo (pHash, pQuery = {}, pForward = 0, pEvent = null) {
 
     // close the command-box when it is stil open
     CommandBox.hideManualRun();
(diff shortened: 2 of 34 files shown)


erwindon:
migrated to PR

erwindon:
@rcmoutinho 
I've turned this issue into a PR so that I can experiment on a branch and let that be visible for you (and others).
The first implementation tries to use the CTRL key to open the page in a new tab.
We'll see how that works out...

rcmoutinho:
Outstanding! Sounds great to me! Looking forward to seeing your tests! Thanks again!

erwindon:
I've pushed a small implementation on the branch from this PR.
* ALT-click now opens a new page in a new browser-tab.
I tried CTRL-click, but that interfered too much with the "select-text" function of the browser.
* For this branch this ONLY works for page "Grains" at this moment:
** For ALT-click on the table-row; and
** For ALT-click on the (only) menu item of the menu for each row.
* The new page has no idea that it was opened this way. But there are several possible variations:
** The new page does not have a logo/menubar/cmdbutton; and/or
** The new page does not have a close-button on the page; and/or
** The new page does not have a "last-7-jobs" panel; and/or
** suggestions?...

rcmoutinho:
To be honest, whatever works to open a new tab, I would be happy.

And just tested locally, and the ALT-Click works great 🎉 

My opinion on what's available on the new page is just what is expected from the click. I would not add a menu or last job panel, just the information about the click (grain, pillar, job, or whatever info we are clicking). It would be a simple light call to the server to collect the precise information it needs. Once we read the page, we can just close it and continue on the context of the main page already opened.

Let me know what you think about it.
Thanks for the awesome improvement!

erwindon:
thx for the feedback!
It may take a while before I have this complete as I now have the following tasks:
* update all table-row (navigation)clicks to allow open in new window
* update all popup-menu (navigation)clicks to allow open in new window
* provide the new tab/window with additional information (in the url) to indicate that it is a new tab/window
* hide the on-page close-button on a new tab/window
* hide the menu bar on a new tab/window (keep the "SaltGUI"-part visible?)
* hide the jobs panel on a new tab/window
* test with logout on main-window, what should happen on/with secondary windows?
* update documentation
* think about a visual clue for this and add that too

rcmoutinho:
Damn! Indeed, it's a lot, but you have a great plan.

Would hiding info prevent sending extra calls to the server or just not showing on the UI?

I am okay with or without the SaltGUI logo/menu. The information is the most essential part. I could execute the same module call to get it, but it would defeat the UI's purpose of making it easier/quicker to get the information.

gseguinbourgeois:
Thank you for making this possible.

This little feature makes SaltGUI become a power tool.
When launching jobs, highstates, orch on dozens of minions at a time, I think this tool should limit the noise to a maximum. (events, find_job, etc.)
And I think this here is the biggest part.

The list of tasks seem good. Good luck with the design as well!

erwindon:
> Would hiding info prevent sending extra calls to the server or just not showing on the UI?

As you have already seen, there is a separation between the pages and the panels.
When you look at the code for the pages, you'll see that it is just the code to create the desired panels and add them.
That code will be updated to only create+add the left-most panel and not create+add the jobs panel.
Api-calls will only take place for the created+added panels.

erwindon:
- [ ] update all table-row (navigation)clicks to allow open in new window
- [ ] update all popup-menu (navigation)clicks to allow open in new window
- [ ] provide the new tab/window with additional information (in the url) to indicate that it is a new tab/window
- [ ] hide the on-page close-button on a new tab/window
- [ ] hide the menu bar on a new tab/window (keep the "SaltGUI"-part visible?)
- [ ] hide the jobs panel on a new tab/window
- [ ] test with logout on main-window, what should happen on/with secondary windows?
- [ ] update documentation
- [ ] think about a visual clue for this and add that too

erwindon:
> I tried CTRL-click, but that interfered too much with the "select-text" function of the browser.

Enabled it anyway.
This has the fun-behavior that CTRL-click opens in a new tab without making that the current tab and ALT-click opens in a new tab which becomes the current tab. Note that that happens automatic, without specific instructions in SaltGUI. So it may behave different in different browsers. But at least Firefox+Chrome+Edge do this.

rcmoutinho:
Interesting! I remember that once I clicked it, it opened and activated the new tab. However, I think the default behavior on the browser for Alt is to "save link as". I will test all your changes first thing Monday morning and give you extra feedback if there is any.

Thanks again for the massive code changes to make it happen 🎉

erwindon:
> fix tooltips at top of window that (with the header removal) fall off the top of the window

solution is to show the header again, only the menu-items are now hidden in secondary tab/page.

rcmoutinho:
This is beautiful! Opening five job results simultaneously using CTRL is even better than ALT (considering it changes the page focus). Both approaches are working on my side 🎉 

I enabled all notifications to ensure I update them as soon as possible once this is merged with the master.
Thanks a lot again! 🤩

erwindon:
@rcmoutinho
I've handled the last two tasks.
The visualization of this all is delegated to a separate issue #575.
I found how to undo the table-cell-selection that happens on a ctrl-click in a table.
This concludes the coding part of this issue/PR.

I plan to make a new release of SaltGUI after this issue and issue #571 are done.

Can you give it one final round of testing?

PR #575, opened 2024-04-30
add a navigation indicator for table rows

erwindon:
clicking on table row has various effects: it may navigate to a details page or it may show the cmd-popup.
for menu items it is less obvious since the description already gives a clue and the "..." shows that a popup will be used.
with the introduction of #574 that becomes less obvious

add a i-icon to each table to show what will happen when you click on a row of that table.

**Additional context**
see #574

### DIFF
--- a/saltgui/static/scripts/panels/Minions.js
+++ b/saltgui/static/scripts/panels/Minions.js
@@ -25,7 +25,7 @@ export class MinionsPanel extends Panel {
     this.addWarningField();
     this.addTable(["Minion", "Status", "Salt version", "OS version", "-menu-"]);
     this.setTableSortable("Minion", "asc");
-    this.setTableClickable();
+    this.setTableClickable("cmd");
     this.addMsg();
   }
 
--- a/saltgui/static/scripts/panels/Beacons.js
+++ b/saltgui/static/scripts/panels/Beacons.js
@@ -13,7 +13,7 @@ export class BeaconsPanel extends Panel {
     this.addSearchButton();
     this.addTable(["Minion", "Status", "Beacons", "-menu-"]);
     this.setTableSortable("Minion", "asc");
-    this.setTableClickable();
+    this.setTableClickable("page");
     this.addMsg();
   }
 
--- a/saltgui/static/scripts/panels/BeaconsMinion.js
+++ b/saltgui/static/scripts/panels/BeaconsMinion.js
@@ -32,7 +32,7 @@ export class BeaconsMinionPanel extends Panel {
     ]);
     this.addTable(["Name", "-menu-", "Config", "Timestamp", "Value", "-help-"]);
     this.setTableSortable("Name", "asc");
-    this.setTableClickable();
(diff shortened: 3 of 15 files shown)


erwindon:
> add a i-icon to each table that (on-hover) shows what will happen when you click on a row of that table

this function is added.
@rcmoutinho FYI

erwindon:
@rcmoutinho
have you tried it?

@everyone
any hints on how to show this for the items op a popup-menu are welcome!

saltgui/static/scripts/panels/Minions.js
```diff
@@ -25,7 +25,7 @@ export class MinionsPanel extends Panel {
     this.addWarningField();
     this.addTable(["Minion", "Status", "Salt version", "OS version", "-menu-"]);
     this.setTableSortable("Minion", "asc");
-    this.setTableClickable();
+    this.setTableClickable("cmd");
```
rcmoutinho, line 28: This page cannot manage ALT or CRTL on a row click. I would remove the info command.

_PS: I assume this is the home page panel, and this code adds the info reference. If this is not the case, ignore this comment or apply it to the correct file._

rcmoutinho:
So far, it works fine for me. I made just one comment so you can double-check on your side.

saltgui/static/scripts/panels/Minions.js
erwindon: thx, I will look into it once I am back at my dev environment...

FYI
The "Minions" page is the default homepage.
Any other page can be configured to be the homepage.
See also https://github.com/erwindon/SaltGUI/#reduced-menus.

erwindon: the function `setTableClickable` has 3 possible parameters:
* `cmd`: when clicking on a row shows the "CMD" popup (and no special handling of CTRL/ALT)
* `page`: when clicking on a row navigates to a new page and considers CTRL/ALT to do that in a new tab
* (none): when clicking on a row does nothing at all.

the presence of a parameter-value (i.e. the first 2 cases) causes the info-icon to be shown in to top-right corner of the table. the information shown there is different for each case.

so, the original code was correct.

erwindon: I verified each occurrence of `setTableClickable` and found no unintended behavior.

The only odd one is the rows from page "Nodegroups" (only visible when you actually have nodegroups defined).
All regular rows properly behave with "cmd". But this table also shows minions that are not in "accepted" state. For these rows, the click is ignored and when not ignored should act using "page". I decided earlier that it better to not do that.

erwindon:
split this in 2 parts, one for table-rows (this one) and for menu items (#579)

saltgui/static/scripts/panels/Minions.js
rcmoutinho: Thanks for the extra details!

erwindon:
I changed the style of the hint from a (?) icon to the style that is also used for panel-related help (but a bit smaller)

PR #579, opened 2024-05-01
add a navigation indicator for menu items

erwindon:
clicking on table row has various effects: it may navigate to a details page or it may show the cmd-popup.
for menu items it is less obvious since the description already gives a clue and the "..." shows that a popup will be used.
with the introduction of #574 that becomes less obvious

* add clue to menuitems to show what will happen when you select that menuitem?

**Additional context**
see #574 and #575


erwindon:
I could not find a proper visualization for this.
The indicator for tables was added, let's hope that is sufficient.

PR #581, opened 2024-05-08
move all in-table menu-items to the first column

erwindon:
in-table dropdown menus currently are mostly at the end of the row. Unless a potentially big last column is present, in which case the menu is moved one column to the second-to last position (possibly third-to-last, etc).
for task items, often the dropdown menu is off-window

Move all in-table dropdown menus to the start of their row.

**Additional context**
This also improves consistency...


erwindon:
visual walkthrough --> ok
code walkthrough --> ok

PR #583, opened 2024-05-11
timers are not always stopped when leaving a page

erwindon:
**Describe the bug**
When leaving a page, the associated timer-sequence continues.
In some cases, returning to the page starts a new timer, causing multiple timers to be active.

**To Reproduce**
tbd

**Expected behaviour**
Page/panel-related timers must be stopped when leaving the page.
See e.g. `options.js`. function `onHide`.

**Additional context**
one positive factor: most timer-sequences are not infinite sequences.

proposal to create panelTimerHelper function.
timer is registered in Panel.
timer is disabled when `isFinished`=true, or when `onHide` is called
parameters:
* actionCallback function
* runFirstTime flag, runs the callback extra immediately
* isFinished function, signals that no more runs are needed
* isSkipped function, signals that this run is not needed
* playPause setting is always considered for skipping the function
* playPause will be hidden when `isFinished`=true

### DIFF
--- a/saltgui/static/scripts/issues/State.js
+++ b/saltgui/static/scripts/issues/State.js
@@ -65,7 +65,7 @@ export class StateIssues extends Issues {
       jobs = jobs.slice(0, MAX_HIGHSTATE_JOBS);
     }
 
-    this.jobs = jobs;
+    pPanel.jobs = jobs;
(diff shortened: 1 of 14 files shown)


erwindon:
no library added, just plain fixes.

PR #605, opened 2024-06-26
saltgui_preview_grains: Improvement

erwindon:
Will be really nice to have better visualization of grains that contain list of values (strings), for example:
image
```
# salt salt-master grains.get roles
salt-master:
    - salt-master
    - salt-minion
```
```
salt-master-custom.conf:
...
saltgui_preview_grains:
  - Roles=roles    
  - Roles2=$.roles.[*]
  - Roles3=$.roles[*]
```

Instead:
```
[
    "salt-master",
    "salt-minion"
]
```
will be nice to have a list separated by `\n` I guess (without brackets and quotes):
```
salt-master
salt-minion
```

### DIFF
--- a/saltgui/static/stylesheets/grains.css
+++ b/saltgui/static/stylesheets/grains.css
@@ -1,5 +1,6 @@
 td.grain-value {
   white-space: pre-wrap;
+  word-break: keep-all;
 }
 
 #page-grains {


erwindon:
> should I create PR for these changes or you will or you think of doing something else?

I just converted this github-issue to a github-pr which is connected to my own branch for this.
when you switch your branch to this one, you should see the same code.
(well almost, because I added a css clause and you replaced the existing one)
I don't want to break anything, so I have a few tests to perform.
that will take a couple of days...

erwindon:
thx for that info!
I'll have to test the many combinations of:
* the locations that use this CSS; and
* the different variations of the possible data; and
* all 3 representations (nested/json/yaml).

erwindon:
FYI: the css class `grain-value` is only used on the grains-overview page and the grains-per-minion page.

> I guess you need to keep only word-break: keep-all; ...

for tight spaces that indeed behaves just a bit better for yaml output.
but it also removes all indentation when using the (default) json output, and that is worse.
so, I'm keeping the current PR unchanged.
do you see any further bad side-effects?

erwindon:
thx!

PR #624, opened 2024-08-11
add proper syntax highlighting for returned data

erwindon:
SaltGui shows (structured) results in a plain way. A few low-level formatting tricks are use to format it.

Use library that can do syntax highlighting on data.

**Additional context**
* library must be free to use
* library must be simple to use
* library must be small (as shipped)


erwindon:
use https://highlightjs.org/

PR #660, opened 2024-12-06
get local test providers for targeted minions only

vkotarov:
This PR reduces the amount of calls to minions when "Manual run" window is opened after a click on a row in the  "Minions" table.

Up until now the behaviour is such that all minions are targeted with "test.providers", no matter what is in the target selection box.

### DIFF
--- a/saltgui/static/scripts/Api.js
+++ b/saltgui/static/scripts/Api.js
@@ -180,11 +180,11 @@ export class API {
     return this.apiRequest("POST", "/", params);
   }
 
-  getLocalTestProviders () {
+  getLocalTestProviders (pTgt) {
     const params = {
       "client": "local",
       "fun": "test.providers",
-      "tgt": "*"
+      "tgt": pTgt || "*"
     };
     return this.apiRequest("POST", "/", params);
   }
--- a/saltgui/static/scripts/Api.js
+++ b/saltgui/static/scripts/Api.js
@@ -180,12 +180,25 @@ export class API {
     return this.apiRequest("POST", "/", params);
   }
 
-  getLocalTestProviders (pTgt) {
+  getLocalTestProviders (pTgt, pTgtType) {
+    pTgtType = pTgtType || 'glob';
+
+    switch (pTgtType) {
+      case 'glob':
+        pTgt = pTgt || '*';
+        break;
+      case 'list':
+        pTgt = pTgt.split(',').map(t => t.trim());
+        break;
+    }
+
     const params = {
       "client": "local",
       "fun": "test.providers",
-      "tgt": pTgt || "*"
+      "tgt": pTgt || "*",
+      "tgt_type": pTgtType || 'glob'
     };
+
     return this.apiRequest("POST", "/", params);
   }
 
--- a/saltgui/static/scripts/CommandBox.js
+++ b/saltgui/static/scripts/CommandBox.js
@@ -477,15 +477,18 @@ export class CommandBox {
     CommandBox._populateTemplateCatMenu();
     CommandBox._populateTemplateTmplMenu();
 
-    targetField.addEventListener('change', (evt) => {
-      const localTestProviders = pApi.getLocalTestProviders(targetField.value);
+    const refreshTestProviders = (evt) => {
+      const targetType = TargetType.menuTargetType._value;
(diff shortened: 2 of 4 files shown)


erwindon:
@vkotarov 

first of all: well spotted, thx!

But I'm sorry, I will not apply this PR. It would work, but an even simpler solution comes in mind.

The reason is that the information that is collected can be assumed to be static. The list of providers that are installed on the minions does not really change. It is therefore sufficient to not even make the call when the required information is already collected before.

The collected information is stored in global variable `Documentation.PROVIDERS`.
A simple test on that variable can be added to determine whether the next call is even needed.

The last 7 lines of function `showManualRun` should be moved to a function.
That function would just immediately return when `Documentation.PROVIDERS` is a non-empty dictionary.

erwindon:
I think PR #661 will do the job too.
Can you please try that one?

dexterlb:
Thank you for the quick reaction!

You are right that subsequent calls for the same information would be unneeded. Unfortunately, the problem in our deployment is not so much that the request is executed every time the menu is opened, but rather that the request itself takes a huge amount of time because it targets all minions (our deployment has thousands of minions).

Furthermore, some minions have a different set of providers available, so I guess a proper solution would have to somehow keep track of the available providers for every target we've requested and not request it a second time?

In general, we're experiencing multiple scalability issues with SaltGUI because of requests that target `'*'` - if you have another suggestion as to how to mitigate this, I am open to hearing it!

erwindon:
The function of collecting the list of providers is only to set the expectation when menu-function "Online reference for 'xxxx'" is used. This option acts on what is already typed in in field "Command". It prevents that SaltGUI presents a documentation link that does not exist.
Since the documentation is a global resource, it makes sense to take the union of all minions to decide what is recognized.

erwindon:
I can add a simple configuration option in SaltGUI helps with this.
So far I can think of the following solutions for that:
# disable the call to `Documentation.PROVIDERS`
The result would be that SaltGUI would offer links to invalid pages.
In the current code, it sets the expectation by warning "The documentation index has not been retrieved yet. We'll just assume this is a regular command.". We may need to slightly adjust that one in this case.
# pre-fill `Documentation.PROVIDERS` with the given providers from the configuration file. problem is that there are 500+ modules. and it would ignore any site-custom modules.
# pre-fill `Documentation.PROVIDERS` with the given providers from the SaltGUI code. problem is again that there are 500+ modules, which is hard to maintain for me. and it would again ignore any site-custom modules.

from this set, only the first one is reasonable.
any thoughts or suggestions?

vkotarov:
@erwindon 

Adding more points to consider in the options above (which actually strengthen option #1) 

1. There might be custom developed modules (we have some of these) loaded on the minions and for them the only way to get some documentation is calling sys.doc as they aren't publicly available
2. Some modules may not be loaded due to failed prerequisites and this is minion-dependent. Showing help for such modules on minions that haven't had them loaded is misleading.

Considering these points, as well as [this comment](https://github.com/erwindon/SaltGUI/pull/661#issuecomment-2531395609), I'd either go for modules list collected on per-minion or have the option to disable this functionality at all.

erwindon:
@vkotarov 

> Some modules may not be loaded due to failed prerequisites and this is minion-dependent. Showing help for such modules on minions that haven't had them loaded is misleading.

The purpose of the call is to reduce the errors on providing documentation.
Specifically avoiding to provide URLs to pages that do not exist.
I think that a per-minion approach is too complex for that.

I will add a variable with which the `test.providers` call can be tuned:
* when not set, it will use the current behavior of using glob target `*`; and
* when set to value "skip", it will not perform that call at all; and
* any other value will be treated as a target-specification (the appropriate target-type will be chosen, `list` when the value contains at least one `,`, `compound` when the value contains at least one `@`, etc.). This makes it possible to point to a few fast and always-on minions.

I'll link to the new PR as soon I have it.

dexterlb:
@erwindon Thank you! This seems like a good compromise.

erwindon:
@vkotarov @dexterlb 
Are you the same person?

See PR #663 for the announced alternative solution.
please discuss that PR only on its own github page (#663)

vkotarov:
closing as #663 provides a better solution.

PR #661, opened 2024-12-07
only call test.providers when results were not retrieved before

erwindon:
general alternative for #660 (but not a replacement for it)

### DIFF
--- a/saltgui/static/scripts/CommandBox.js
+++ b/saltgui/static/scripts/CommandBox.js
@@ -476,6 +476,15 @@ export class CommandBox {
 
     CommandBox._populateTemplateCatMenu();
     CommandBox._populateTemplateTmplMenu();
+    CommandBox._populateProviders(pApi);
+  }
+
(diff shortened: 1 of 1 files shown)


erwindon:
@vkotarov
Can you test this PR to see whether it works also in your case?
We'll discuss further measures in #660, but this is a generic improvement.

vkotarov:
@erwindon 
The thing is that "test.providers" targeting "*" in our environment takes about 85 seconds to complete.
Also, if a user hits the refresh button for some reason, this will have to be executed again, so 85 more seconds time to wait.

erwindon:
@vkotarov 
I know! And that is why the work on #660 will continue. But this one is a general improvement which does not need any configuration. It is an improvement compared to any previous version and it is useful for everyone. See this as a spin-off.
Since it is related to #660, I'm asking you to please look at it and maybe test it in any environment you have available and confirm that it is suitable for the world. thx!

PR #700, opened 2025-05-29
SaltGUI slowness

erwindon:
Hi Team,

I have a Salt Master running on Ubuntu 22.04 with Salt version 3006.1, managing over 1000 minions. To simplify management, I’ve configured SaltGUI for access. However, I’m noticing significant performance issues — particularly, it takes a long time to load and display the list of minions after logging into SaltGUI.

The Salt Master is running on a server with a 16-core CPU and 12 GB of RAM. I’d like to know if this hardware specification is sufficient for managing this number of minions, or if an upgrade is recommended. Additionally, are there any specific configuration best practices for optimizing Salt/SaltGUI performance in large environments like this?

Looking forward to your suggestions.

Best regards,
Sunder

### DIFF
--- a/saltgui/static/scripts/CommandBox.js
+++ b/saltgui/static/scripts/CommandBox.js
@@ -459,17 +459,6 @@ export class CommandBox {
       targetList.appendChild(option);
     }
 
-    if (Utils.getStorageItem("session", "minions", "N/A") === "N/A") {
-      const wheelKeyListAllPromise = pApi.getWheelKeyListAll();
-      /* eslint-disable no-unused-vars */
-      wheelKeyListAllPromise.then((pWheelKeyListAllData) => {
-        const minions = pWheelKeyListAllData.return[0].data.return.minions;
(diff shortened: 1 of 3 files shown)


erwindon:
I'm keeping this issue open to look at the list of minions to be available in de CommandBox under more/all cicumstances.

erwindon:
@sunderjayaraj 
SaltGUI branch `cmd_minion_list` now contains a crude solution for this.
The CommandBox will collect the information if it was not collected before.
The downside is that (for now) the result is only visible when you open the CommandBox for the second time.
I'll look whether that can be improved.
Can you please try out the branch from this PR?

Note that I've made fixes and merges for other issues. Please use `git pull -r` when updating that branch...

erwindon:
ping...

erwindon:
I've moved the code to initialize this list to always run on login.
This makes that procedure uniform as it is now always run on login only.
It will therefore work independent of the menu-bar authorizations.

@sunderjayaraj I will merge this code as it is an improvement, thx for bringing it up.

PR #871, opened 2026-04-18
Fix typo /src to /srv

clairem-sl:
Just a simple typo fix.


erwindon:
there is nothing worse than documentation that does not tell the thruth.
thx!

PR #974, opened 2026-09-16
Add validation framework with fast visual feedback

erwindon:
* works on the CommandBox
* validations show indications for what they find
* separation between warnings and errors, warnings do not prevent execution of the command
* one such warning is "integer overflow", as suggested by @rawsun007


rawsun007:
Thanks for the heads-up. BigInt then range check is the right shape - it avoids the thing that made #964 awkward, which is that the expected values in the existing assertions are themselves JS numbers and so round the same way the input does.

On the red quality gate here, since I spent two rounds on the same one: it is entirely the test preamble, not the framework code. SonarCloud's API gives the breakdown without opening the dashboard:

```
new_duplicated_lines = 332 of 1100 new lines  (30.2%)
  tests/unit/ParseCommandLine.test.js   185
  tests/unit/CommandBox.test.js         111
  saltgui/static/scripts/CommandBox.js   36
```

The largest cluster is one 14-line shape repeated thirteen times in `ParseCommandLine.test.js` - `args = []; params = {}; tokens = [];`, the `parseCommandLine(...)` call, then the same four assertions. Blocks at 39-52, 166-179, 176-191, 188-201, 208-221, 218-231, 338-350, 347-359, 600-614, 611-623, 629-644, 641-655, 679-691.

What cleared it on #964 was turning those into a table and looping: one `for (const [input, expectedArg] of [...])` covering the same inputs took the six cells I had added from 39% to under the threshold, with no loss of coverage. The gate measures *new* code only, so inheriting the file's existing style is what trips it.

One other thing from that round, in case it bites here too: my second red gate was a reliability rating, not duplication, and it was `assert.equal(String(args[0]), nr, nr)` - the same variable passed as both the expected value and the failure message, which Sonar reads as a copy-paste slip. `curl -s "https://sonarcloud.io/api/issues/search?componentKeys=erwindon_SaltGUI&pullRequest=974&resolved=false"` lists those with file and line; it reports 0 for this PR right now, so only the duplication stands between it and a green gate.

Happy to send the table-driven rewrite of `ParseCommandLine.test.js` against your branch if that saves you a pass - it is mechanical and it is the file I already did this to.

Written with Claude Opus 5 in Claude Code, under my account; the numbers above come from the SonarCloud API for this PR.

erwindon:
> new_duplicated_lines = 332 of 1100 new lines  (30.2%)

I noticed that too, but since the duplication is mostly in the unit tests (185+111=296 of 332), I did not care. I've created a new issue (#975) to remind myself to reduce this.

> assert.equal(String(args[0]), nr, nr)

where did you find that?

> Happy to send the table-driven rewrite of ParseCommandLine.test.js against your branch if that saves you a pass - it is mechanical and it is the file I already did this to.

yes, would be nice to see. can you add it to #975?

